PR #12, opened 2017-10-07
Use interpolated tags syntax to add extra tags to cluster nodes

sclausson:
https://github.com/hashicorp/terraform-aws-vault/issues/7

### DIFF
--- a/modules/vault-cluster/variables.tf
+++ b/modules/vault-cluster/variables.tf
@@ -80,6 +80,20 @@ variable "cluster_tag_key" {
   default     = "Name"
 }
 
+variable "cluster_extra_tags" {
+  description = "A list of additional tags to add to each Instance in the ASG. Each element in the list must be a map with the keys key, value, and propagate_at_launch"
+  type = "list"
+  #example: 
+  # default = [
+  #   {
+  #     key = "Environment"
+  #     value = "Dev"
+  #     propagate_at_launch = true
+  #   } 
+  # ]
+  default = []
+}
+
 variable "termination_policies" {
   description = "A list of policies to decide how the instances in the auto scale group should be terminated. The allowed values are OldestInstance, NewestInstance, OldestLaunchConfiguration, ClosestToNextInstanceHour, Default."
   default     = "Default"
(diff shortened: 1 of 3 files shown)


modules/vault-cluster/variables.tf
```diff
@@ -76,8 +76,13 @@ variable "allowed_ssh_security_group_ids" {
...
 variable "cluster_tag_key" {
-  description = "Add a tag with this key and the value var.cluster_name to each Instance in the ASG."
-  default     = "Name"
+  description = "Add a tag with this key and the value var.cluster_tag_value to each Instance in the ASG."
+  default     = "vault-servers"
+}
+
+variable "cluster_tag_value" {
+  description = "Add a tag with this value and the key var.cluster_tag_key to each Instance in the ASG."
+  default     = "auto-join"
```
brikis98: Looking at this with fresh eyes, it seems like the approach we took with tags is a little too hard-coded. A better approach may be to take advantage of the [tags](https://www.terraform.io/docs/providers/aws/r/autoscaling_group.html#tags) parameter on ASGs. Something along the lines of:

```hcl
variable "tags" {
  type = "list"
  # Example:
  # default = {
  #   {
  #     key = "foo"
  #     value = "bar"
  #     propagate_at_launch = true  
  #   }
  # }
}

resource "aws_autoscaling_group" "vault" {
  # ...
  tags = "${merge(map("Name", var.cluster_name), var.tags)}"
}
```

This way, users can set any arbitrary tags they want, including overriding the `Name` if they so choose.

sclausson: @brikis98 I agree this is a much better solution. I'll made the changes in my PR.

sclausson:
@brikis98 I've amended this commit to support the interpolated tags syntax.

modules/vault-cluster/variables.tf
```diff
@@ -80,6 +80,12 @@ variable "cluster_tag_key" {
   default     = "Name"
 }
 
+variable "cluster_extra_tags" {
+  description = "A list of additional tags to add to each Instance in the ASG."
+  type = "list"
+  default = []
```
brikis98: Could you update the `description` to mention that each element in the list should be a map with the keys `key`, `value`, and `propagate_at_launch`? Also, an example in comments above `default` would be helpful.

sclausson:
Ok should be all set now with requested changes.

brikis98:
https://github.com/hashicorp/terraform-aws-vault/releases/tag/v0.0.6

Issue #14, opened 2017-10-08
Wrong var referenced to allow inbound SSH security group IDs


sclausson:
```hcl
resource "aws_security_group_rule" "allow_ssh_inbound_from_security_group_ids" {
  count                    = "${length(var.allowed_inbound_security_group_ids)}"
  type                     = "ingress"
  from_port                = "${var.ssh_port}"
  to_port                  = "${var.ssh_port}"
  protocol                 = "tcp"
  source_security_group_id = "${element(var.allowed_inbound_security_group_ids, count.index)}"

  security_group_id = "${aws_security_group.lc_security_group.id}"
}
```
This resource needs to reference `var.allowed_ssh_security_group_ids` instead of `var.allowed_inbound_security_group_ids`

brikis98:
Ah, you're right!

PR #15, opened 2017-10-08
Fix incorrect var reference in inbound ssh sg ids

sclausson:
Fixes https://github.com/hashicorp/terraform-aws-vault/issues/14


brikis98:
https://github.com/hashicorp/terraform-aws-vault/releases/tag/v0.0.4

Issue #22, opened 2017-11-01
Add Optional IAM Role Name Variable


jeffwecan:
As is, the `aws_iam_role ` resource created in the vault-cluster module defaults to using the `cluster_name` variable as the role resource's `name_prefix` argument (https://github.com/hashicorp/terraform-aws-vault/blob/v0.0.7/modules/vault-cluster/main.tf#L158-L159). As a result, the ultimate instance role ARN can't be determined statically. It would be helpful in my use case to have an optional `iam_role_arn` variable to ensure the same ARN across deployments. This variable could have a default that sees the current `name_prefix` value maintained. Then, if the new variable was set to a different value, `name` would be used in place of `name_prefix` on the `aws_iam_role` resource.

For context, the use case I am envisioning is role-based access provided to the vault nodes with the associated policy managed outside of this vault_cluster module. I.e., I need to provide the vault instance role ARN to another codebase and ideally have that ARN remain constant across vault deployments.

brikis98:
The module does return the IAM ARN info as output variables, so you shouldn't have to hard-code them: https://github.com/hashicorp/terraform-aws-vault/blob/master/modules/vault-cluster/outputs.tf#L21-L27. 

That said, this feature would likely be useful for a number of reasons, so I'm in favor of it. It should be straightforward to add an `iam_role_name` input variable, set to an empty string by default, and to name the IAM role using that variable if it's non-empty or `var.name` otherwise.

PRs welcome!

jeffwecan:
I'm definitely down with using the module's output variables to programmatically get at the role ARN when appropriate. To clarify the issue's description:

* It isn't a matter of the role name being unknown post-deploy, its a matter of the ARN not being known *before* invoking `terraform apply`. 
* Additionally, it is conceivable that the role ARN would change on subsequent deployments. For instance, if the cluster_name variable was modified.

That said, this issue was filed to ensure the idea wasn't completely misguided (and to abide by per the repo's contributing guidelines). I totally intend to try my hand at a PR for this fella! :sweat_smile:

PR #23, opened 2017-11-02
Add Optional IAM Role Name Variable

jeffwecan:
Associated `terraform-aws-vault` Module Issue: https://github.com/hashicorp/terraform-aws-vault/issues/22

This changeset is the best I can muster as a terraform novice; so I naturally welcome any feedback on terraform best practices / preferred style. 

Since the `aws_iam_role` resource's `name` and `name_prefix` arguments are conflicting arguments (you may provide one but not both), This is one of those funny cases where it would be kinda swell if terraform allowed some sort of conditional logic. Props to @alexsomesan and [this upstream issue comment](https://github.com/hashicorp/terraform/issues/11566#issuecomment-289417805) for providing an example of the pattern at the core (😜)  of the solution proposed here.

No test updates included. I considered if adding testing for the case of a user-provided value of the newly added iam_role_name variable. However that feels like testing how terraform itself behaves rather than the responsibility of this codebase(?).

### DIFF
--- a/modules/vault-cluster/main.tf
+++ b/modules/vault-cluster/main.tf
@@ -145,7 +145,7 @@ module "security_group_rules" {
 resource "aws_iam_instance_profile" "instance_profile" {
   name_prefix = "${var.cluster_name}"
   path        = "${var.instance_profile_path}"
-  role        = "${aws_iam_role.instance_role.name}"
+  role        = "${data.aws_iam_role.instance_role.name}"
 
   # aws_launch_configuration.launch_configuration in this module sets create_before_destroy to true, which means
   # everything it depends on, including this resource, must set it as well, or you'll get cyclic dependency errors
@@ -155,9 +155,12 @@ resource "aws_iam_instance_profile" "instance_profile" {
   }
 }
 
-resource "aws_iam_role" "instance_role" {
-  name_prefix        = "${var.cluster_name}"
-  assume_role_policy = "${data.aws_iam_policy_document.instance_role.json}"
+resource "aws_iam_role" "prefixed_instance_role" {
+  # If the iam_role_name variable contains an empty string, use the cluster_name variable for this resource's
+  # name_prefix argument.
+  count               = "${var.iam_role_name == "" ? 1 : 0}"
+  name_prefix         = "${var.cluster_name}"
+  assume_role_policy  = "${data.aws_iam_policy_document.instance_role.json}"
 
   # aws_iam_instance_profile.instance_profile in this module sets create_before_destroy to true, which means
   # everything it depends on, including this resource, must set it as well, or you'll get cyclic dependency errors
@@ -165,6 +168,26 @@ resource "aws_iam_role" "instance_role" {
   lifecycle {
     create_before_destroy = true
   }
+}
+
+resource "aws_iam_role" "named_instance_role" {
+  # If the iam_role_name variable is set, use its value for this resource's name argument.
+  count               = "${var.iam_role_name != "" ? 1 : 0}"
+  name                = "${var.iam_role_name}"
+  assume_role_policy  = "${data.aws_iam_policy_document.instance_role.json}"
+
+  # aws_iam_instance_profile.instance_profile in this module sets create_before_destroy to true, which means
+  # everything it depends on, including this resource, must set it as well, or you'll get cyclic dependency errors
+  # when you try to do a terraform destroy.
+  lifecycle {
+    create_before_destroy = true
+  }
+}
+
+data "aws_iam_role" "instance_role" {
+  # The join() hack is required because currently the ternary operator evaluates the expressions on both branches of the
+  # condition before returning a value. This is tracked upstream: https://github.com/hashicorp/hil/issues/50
+  name = "${var.iam_role_name == "" ? join(" ", aws_iam_role.prefixed_instance_role.*.name) : var.iam_role_name }"
 }
 
 data "aws_iam_policy_document" "instance_role" {
@@ -195,7 +218,7 @@ resource "aws_s3_bucket" "vault_storage" {
 
 resource "aws_iam_role_policy" "vault_s3" {
   name   = "vault_s3"
-  role   = "${aws_iam_role.instance_role.id}"
+  role   = "${data.aws_iam_role.instance_role.id}"
   policy = "${data.aws_iam_policy_document.vault_s3.json}"
 }
 
--- a/modules/vault-cluster/main.tf
+++ b/modules/vault-cluster/main.tf
@@ -145,7 +145,7 @@ module "security_group_rules" {
 resource "aws_iam_instance_profile" "instance_profile" {
   name_prefix = "${var.cluster_name}"
   path        = "${var.instance_profile_path}"
-  role        = "${data.aws_iam_role.instance_role.name}"
+  role        = "${data.template_file.instance_role_name.rendered}"
 
   # aws_launch_configuration.launch_configuration in this module sets create_before_destroy to true, which means
   # everything it depends on, including this resource, must set it as well, or you'll get cyclic dependency errors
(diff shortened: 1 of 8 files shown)


modules/vault-cluster/variables.tf
```diff
@@ -156,6 +156,11 @@ variable "health_check_grace_period" {
   default     = 300
 }
 
+variable "iam_role_name" {
+  description = "The name to use for the Vault instances' IAM role. Use this setting to ensure a determensitc IAM role ARN, otherwise the default will see terraform create a unique IAM role name (prefixed by the vault_cluster_name variable)."
```
brikis98: s/determensitc/deterministic/

brikis98: This applies to all the `variables.tf` files that were updated in this PR.

modules/vault-cluster/main.tf
```diff
@@ -167,6 +170,26 @@ resource "aws_iam_role" "instance_role" {
...
+  lifecycle {
+    create_before_destroy = true
+  }
+}
+
+data "aws_iam_role" "instance_role" {
+  # The join() hack is required because currently the ternary operator evaluates the expressions on both branches of the
+  # condition before returning a value. This is tracked upstream: https://github.com/hashicorp/hil/issues/50
+  name = "${var.iam_role_name == "" ? join(" ", aws_iam_role.prefixed_instance_role.*.name) : var.iam_role_name }"
+}
```
brikis98: Hm, I suspect this will be slightly problematic, as it doesn't explicitly depend on the `aws_iam_role.named_instance_role` resource, so this data source may try to fetch that instance role before it's actually created. Also, the extra data source (and extra API call) seems unnecessary. I think you can replace it with a `template_file` instead:

```hcl
data "template_file" "instance_role_name" {
  template = "${element(concat(aws_iam_role.prefixed_instance_role.*.name, aws_aim_role.named_instance_role.*.name), 0)}"
}
```

You can then use `data.template_file.instance_role_name.rendered` elsewhere in the code.

jeffwecan: Fwiw your suspicions are confirmed, a first-run of an apply with this update would always lead to:

```
* module.vault_cluster.aws_iam_instance_profile.instance_profile: Resource 'data.aws_iam_role.instance_role' not found for variable 'data.aws_iam_role.instance_role.name'
```

I suppose that is why all the test harnessing in these modules actually goes out and creates real resources. 😄  

In hindsight, the pattern / kludge I was using initially is for cases where you want to optionally create a resource if someone isn't bringing their own *pre existing* resource identified in via a var.

jeffwecan:
@brikis98: I gave your suggestion a go and I do think that it gets us most of the way to a satisfactory solution. However there is at least one semi-large gap outstanding around setting the iam_role_name variable and then subsequently reverting back to the default. I haven't been able to determine the exact cause after a good bit of putzing around, so I figured I would go ahead and push up the current state and solicit suggestions. Since I don't have the root cause, I'll just describe the replication steps / behavior observed on this branch:

1. Create a new module that contains the vault_cluster, with the `iam_role_name` left with the default empty string value, plus the [consul-iam-policies module](https://github.com/hashicorp/terraform-aws-consul/tree/v0.0.5/modules/consul-iam-policies) with the consul module's `iam_role_id` config set to `"${module.vault_cluster.iam_role_id}"`. Basically the same setup as [examples/vault-cluster-private/main.tf](https://github.com/hashicorp/terraform-aws-vault/blob/master/examples/vault-cluster-private/main.tf).
2. Run a terraform apply => terraform creates a vault IAM role using the name_prefix argument
3. Update the vault_cluster module config setting `iam_role_name` to some non-empty string
4. Run a terraform apply
5. Terraform destroys the `aws_iam_role_policy.vault_s3` and the `module.consul_iam_policies_servers.aws_iam_role_policy.auto_discover_cluster` resources that were added as an inline policy for the name_prefix role during step 2
6. Terraform creates the new named role
7. Terraform refreshes the state of "module.vault_cluster.data.template_file.instance_role_id" 
8. Terraform **recreates** the `module.consul_iam_policies_servers.aws_iam_role_policy.auto_discover_cluster` resource as an inline policy attached to the **original** name_prefix IAM role
9. Terraform attempts to delete the name_prefix IAM role but fails because it has policies attached:
  ```
    Error applying plan:

    1 error(s) occurred:

    * module.vault_cluster.aws_iam_role.prefixed_instance_role (destroy): 1 error(s) occurred:

    * aws_iam_role.prefixed_instance_role: DeleteConflict: Cannot delete entity, must delete policies first.
    	status code: 409, request id: 54cb8cfd-bff8-11e7-bb63-0bb6f15a9a86
  ```

This only occurs when moving between the two IAM roles on an existing cluster so it is not the worst issue. However it is certainly an unacceptable burden to place on an end-user should the situation arise. The other interesting fact to note is it only occurs when moving from the name_prefixed role  to the named role, not the other way around. So it maybe has something to do with the order of the arguments in our template_file data providers? Otherwise I assume its a similar issue that led me to add `depends_on` argument to the `aws_iam_role_policy.vault_s3` here; terraform is unable to automatically discover that module_consul depends on named_instance_role since we're no longer using that resource directly. 

Here is a gist with some sample output relevant to the scenario described above (for reference the existing name prefix role ended up being called `terraform-vault20171102180130308500000001` and the subsequent value used for `iam_role_name` was `hulloimmavault`): https://gist.github.com/jeffwecan/f38b1b9e6d20e96d6cebfe36eb588f9b.

modules/vault-cluster/main.tf
```diff
@@ -145,19 +145,26 @@ module "security_group_rules" {
...
   # aws_launch_configuration.launch_configuration in this module sets create_before_destroy to true, which means
   # everything it depends on, including this resource, must set it as well, or you'll get cyclic dependency errors
   # when you try to do a terraform destroy.
   lifecycle {
     create_before_destroy = true
   }
+
+  # We must explictly inform terraform of the following dependencies as terraform is unable to resolve them automatically
+  # with a simple string being passed to this resource's role argument (as opposed to an actaul aws_iam_role resource)
+  depends_on = ["aws_iam_role.prefixed_instance_role", "aws_iam_role.named_instance_role"]
```
brikis98: Hm, the dependency on `data.template_file.instance_role_name` should be enough, since that data source already depends on the two `aws_iam_role` resources.

modules/vault-cluster/main.tf
```diff
@@ -195,8 +230,12 @@ resource "aws_s3_bucket" "vault_storage" {
 
 resource "aws_iam_role_policy" "vault_s3" {
   name   = "vault_s3"
-  role   = "${aws_iam_role.instance_role.id}"
+  role   = "${data.template_file.instance_role_name.rendered}"
   policy = "${data.aws_iam_policy_document.vault_s3.json}"
+
+  # We must explictly inform terraform of the following dependencies as terraform is unable to resolve them automatically
+  # with a simple string being passed to this resource's role argument (as opposed to an actaul aws_iam_role resource)
+  depends_on = ["aws_iam_role.prefixed_instance_role", "aws_iam_role.named_instance_role"]
```
brikis98: Same here

brikis98:
The rename issue is likely a bug in Terraform itself: https://github.com/hashicorp/terraform/issues/2761

sarkis:
It's now possible to do this with TF 0.12+:

```
resource "aws_iam_role" "instance_role" {
  name                        = var.iam_role_name ? var.role_name : null
  name_prefix             = var.iam_role_name ? null : var.cluster_name
  assume_role_policy = data.aws_iam_policy_document.instance_role.json

  # aws_iam_instance_profile.instance_profile in this module sets create_before_destroy to true, which means
  # everything it depends on, including this resource, must set it as well, or you'll get cyclic dependency errors
  # when you try to do a terraform destroy.
  lifecycle {
    create_before_destroy = true
  }
}
```

brikis98:
New PR is welcome!

PR #26, opened 2017-11-28
Facilitate specifying non-default VPC (shamelessly ripped off from hashicorp/terraform-aws-consul)

boldandbusted:
Please let me know if you have troubles. I tested this with my local infrastructure (I actually needed it to get past not having a default VPC, which is "AWS Best Practice", at least today. ;) ).

### DIFF
--- a/examples/vault-cluster-private/main.tf
+++ b/examples/vault-cluster-private/main.tf
@@ -123,7 +123,8 @@ data "template_file" "user_data_consul" {
 # ---------------------------------------------------------------------------------------------------------------------
 
 data "aws_vpc" "default" {
-  default = true
+  default = "${var.vpc_id == "" ? true : false}"
+  id = "${var.vpc_id}"
 }
 
 data "aws_subnet_ids" "default" {
--- a/examples/vault-cluster-private/variables.tf
+++ b/examples/vault-cluster-private/variables.tf
@@ -72,3 +72,8 @@ variable "force_destroy_s3_bucket" {
   description = "If you set this to true, when you run terraform destroy, this tells Terraform to delete all the objects in the S3 bucket used for backend storage. You should NOT set this to true in production or you risk losing all your data! This property is only here so automated tests of this module can clean up after themselves."
   default     = false
 }
+
+variable "vpc_id" {
+    description = "The ID of the VPC in which the nodes will be deployed.  Uses default VPC if not supplied."
+    default = ""
+}
--- a/main.tf
+++ b/main.tf
@@ -202,7 +202,8 @@ data "template_file" "user_data_consul" {
 # ---------------------------------------------------------------------------------------------------------------------
 
 data "aws_vpc" "default" {
-  default = true
+  default = "${var.vpc_id == "" ? true : false}"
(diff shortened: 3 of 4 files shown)


examples/vault-cluster-private/main.tf
```diff
@@ -123,9 +123,10 @@ data "template_file" "user_data_consul" {
 # ---------------------------------------------------------------------------------------------------------------------
 
 data "aws_vpc" "default" {
-  default = true
+  default = "${var.vpc_id == "" ? true : false}"
+  id = "${var.vpc_id}"
```
brikis98: Hm, I'm worried what happens if `id` is set to an empty string. Does Terraform ignore it, or does it complain? Unfortunately, Terraform is [inconsistent with how it treats empty vs omitted params](https://github.com/hashicorp/terraform/issues/14037), so I'm not sure if this will cause problems for people trying to use the default VPC. Did you try it that way?

boldandbusted: No, I didn't try it that way - good catch. I'd have to create a default VPC, which I'm not sure is a reversible process in a single AWS account.

boldandbusted:
Re: "examples not production-ready", I get that. However, the main page for this module (as listed in the TF Registry, which is supposed to be production-ready code), refers to the /examples/ directories for working code. Then, in this example's README, it has the NOTE which, to my reading, completely negates what the example code is actually supposed to do (create a private Vault, without public EIPs) and its own description a few paragraphs above it. So, I'm a little confused.

examples/vault-cluster-private/main.tf
brikis98: http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/default-vpc.html#create-default-vpc

brikis98:
> as listed in the TF Registry, which is supposed to be production-ready code

The code in the `modules` folder is production-ready. The code in the `examples` folder, and the root—which is *also* an example—is not necessarily production ready.

examples/vault-cluster-private/main.tf
boldandbusted: Hmm. I'm not comfortable changing my present employer's setup to create and delete a default VPC, since it seems to indicate a bunch of other things are touched by that process (subnets, security groups, etc.), and I don't know the implications of that (or the effect on other Terraform users). Is there another way to test this, perhaps with a free-tier?

boldandbusted:
Got it, and the version, 0.0.8, also indicates it's still for development use. Just FYI, if you naively read the docs as presented on the Registry, it doesn't indicate that the root module is an "example". A blurb in the README.md might help clarify.

examples/vault-cluster-private/main.tf
brikis98: Sure, create a brand new AWS account with the free tier and it'll have a default VPC built-in.

brikis98:
> Just FYI, if you naively read the docs as presented on the Registry, it doesn't indicate that the root module is an "example".

Agreed. I believe the HashiCorp folks have work planned to make the separation between "examples" and prod-ready modules a bit clearer. /cc @KFishner

PR #30, opened 2018-01-02
fixes for consul references in vault cluster example. also upgrades referenced modules from v0.02 to v0.1.0

ugoenyioha:
this was causing terraform init to fail for me since I don't use ssh keys. 

the new format is also referenced in your documentation here. 

https://www.terraform.io/docs/modules/sources.html

as this is an example, I'm going to assume that the added security of an ssh privkey pull isn't a must have.

### DIFF
--- a/examples/vault-cluster-private/main.tf
+++ b/examples/vault-cluster-private/main.tf
@@ -51,7 +51,7 @@ module "vault_cluster" {
 # ---------------------------------------------------------------------------------------------------------------------
 
 module "consul_iam_policies_servers" {
-  source = "git::git@github.com:hashicorp/terraform-aws-consul.git//modules/consul-iam-policies?ref=v0.0.2"
+  source = "github.com/hashicorp/terraform-aws-consul.git//modules/consul-iam-policies?ref=v0.1.0"
 
   iam_role_id = "${module.vault_cluster.iam_role_id}"
 }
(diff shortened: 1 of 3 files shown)


examples/vault-cluster-private/main.tf
```diff
@@ -123,7 +123,8 @@ data "template_file" "user_data_consul" {
 # ---------------------------------------------------------------------------------------------------------------------
 
 data "aws_vpc" "default" {
-  default = true
+  default = "${var.vpc_id == "" ? true : false}"
+  id      = "${var.vpc_id}"
```
brikis98: What is this change?

examples/vault-cluster-private/outputs.tf
```diff
@@ -56,4 +56,28 @@ output "ssh_key_name" {
...
+  value = "${module.consul_cluster.security_group_id}"
+}
+
+output "consul_cluster_cluster_tag_key" {
+  value = "${module.consul_cluster.cluster_tag_key}"
+}
+
+output "consul_cluster_cluster_tag_value" {
+  value = "${module.consul_cluster.cluster_tag_value}"
+}
```
brikis98: What are all these new outputs?

examples/vault-cluster-private/main.tf
ugoenyioha: see: https://github.com/hashicorp/terraform-aws-consul/blob/master/main.tf#L155

what it does is allow you either specify explicitly an id or default to a randomly created one.

examples/vault-cluster-private/outputs.tf
ugoenyioha: see: https://github.com/hashicorp/terraform-aws-consul/blob/master/outputs.tf

these came from the consul project. brings the outputs to parity with that one.

examples/vault-cluster-private/main.tf
ugoenyioha: changed readme and comments. see below.

PR #31, opened 2018-01-03
use splat syntax in fully_qualified_domain_name output

vlfig:
Terraform >0.11.1 warns about non-splat references to resources with `count` defined.

This will become fatal in 0.12 (https://github.com/hashicorp/terraform/pull/16735)

See https://github.com/hashicorp/terraform/blob/v0.11.1/CHANGELOG.md

### DIFF
--- a/modules/vault-elb/outputs.tf
+++ b/modules/vault-elb/outputs.tf
@@ -15,5 +15,5 @@ output "load_balancer_security_group_id" {
 }
 
 output "fully_qualified_domain_name" {
-  value = "${aws_route53_record.vault_elb.fqdn}"
+  value = "${aws_route53_record.vault_elb.*.fqdn}"
 }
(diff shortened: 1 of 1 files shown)


modules/vault-elb/outputs.tf
```diff
@@ -15,5 +15,5 @@ output "load_balancer_security_group_id" {
 }
 
 output "fully_qualified_domain_name" {
-  value = "${aws_route53_record.vault_elb.fqdn}"
+  value = "${aws_route53_record.vault_elb.*.fqdn}"
```
brikis98: This will turn this output into a list, which is probably not what we want. You instead need to use the slightly ugly workaround [listed here](https://www.terraform.io/upgrade-guides/0-11.html#referencing-attributes-from-resources-with-count-0):

```hcl
value = "${element(concat(aws_route53_record.vault_elb.*.fqdn, list("")), 0)}"
```

vlfig: Good point. Wasn't aware of that guide. Have a look.

brikis98:
https://github.com/hashicorp/terraform-aws-vault/releases/tag/v0.0.9

PR #35, opened 2018-01-25
Remove S3

brikis98:
Switch to Consul as both the storage and HA backend. This reduces the
number of dependencies and is the recommended architecture from
HashiCorp.


josh-padnick:
The code itself looks good, but don't we want to support *either* Consul or S3 as the backend?

brikis98:
It was a lot quicker to remove it than to support either option (or even both options). I'm not against adding flags to optionally bring it back, but don't have the time to add that—and more importantly, the example code + tests + docs for it—at the moment.

patoarvizu:
Is there a link to HashiCorp documentation regarding the recommended architecture? I couldn't find it anywhere, at least I couldn't find evidence that the recommendation is to have Consul be BOTH the storage and HA backend.

KFishner:
hey @patoarvizu the main documentation is here: https://www.vaultproject.io/docs/concepts/ha.html

Some key quotes: 
> Vault supports a multi-server mode for high availability. This mode protects against outages by running multiple Vault servers. High availability mode is automatically enabled when using a data store that supports it.

> Certain storage backends can support high availability mode, which enable them to store both Vault's information in addition to the HA lock. However, Vault also supports split data/HA mode, whereby the lock value and the rest of the data live separately. 

> The Consul backend is the recommended HA backend, as it is used in production by HashiCorp and its customers with commercial support.

patoarvizu:
> The Consul backend is the recommended HA backend

The way I interpret that is that Consul is the recommended **HA backend** and not necessarily that Consul is recommended for **both** storage and HA backend. Now, if you're telling me that's the case, I'll believe you but I'd suggest the documentation is updated clarify that :)

I understand that this is done to reduce the complexity and number of dependencies but I'm afraid it does so at the expense of sacrificing durability. Are there any plans to add S3 (as an optional storage backend) back?

KFishner:
> Consul is recommended for both storage and HA backend.
This is correct, agreed the documentation isn't the most clear!

> Are there any plans to add S3 (as an optional storage backend) back?
I assume you mean for this module specifically. If so, there aren't immediate plans to do it based on @brikis98's comment above. Of course S3 is still a valid storage backend for Vault if you're not using this module specifically

brikis98:
> I assume you mean for this module specifically. If so, there aren't immediate plans to do it based on @brikis98's comment above. Of course S3 is still a valid storage backend for Vault if you're not using this module specifically

I will add that:

1. You can use S3 with this module too! Just create the bucket and follow these docs to override the backend configuration: https://github.com/hashicorp/terraform-aws-vault/tree/master/modules/run-vault#overriding-the-configuration.

1. A PR that adds back "official" support for S3 (and example code for it) is very welcome :)

patoarvizu:
Thank you very much! I will consider making a PR :)

PR #37, opened 2018-02-01
Fix permissions on vault folders, cert files

jasonmcintosh:
Vault folders when created with a 600 permissions (umask 0077) don't work when the default user tries to do a setcap command.  Further, TLS files can't have their permissions set via a /opt/vault/tls/* - wild cards don't work with chown when the folder isn't visibly to the requested user.  However, doing a recursive chown works as it doesn't need traversal permissions.

### DIFF
--- a/examples/vault-consul-ami/vault-consul.json
+++ b/examples/vault-consul-ami/vault-consul.json
@@ -77,8 +77,9 @@
       "sudo mv /tmp/ca.crt.pem /opt/vault/tls/",
       "sudo mv /tmp/vault.crt.pem /opt/vault/tls/",
       "sudo mv /tmp/vault.key.pem /opt/vault/tls/",
-      "sudo chown vault:vault /opt/vault/tls/*",
-      "sudo chmod 600 /opt/vault/tls/*",
+      "sudo chown -R vault:vault /opt/vault/tls/",
+      "sudo chmod -R 600 /opt/vault/tls",
+      "sudo chmod 700 /opt/vault/tls",
       "sudo /tmp/terraform-aws-vault/modules/update-certificate-store/update-certificate-store --cert-file-path /opt/vault/tls/ca.crt.pem"
     ]
   },{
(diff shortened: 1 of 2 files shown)


jasonmcintosh:
Tested by building the ami repeatedly debugging repeat.  Though I do need to test on more distributions.  

Issue in this case is the base image has a more restrictive umask causing files and folders to be created with a more restrictive policy.  This umask setting is a CIS hardening recommendation for securing servers 

--
Jason McIntosh 
Infrastructure architect 
Carfax, Inc
573-424-7612

brikis98:
> Tested by building the ami repeatedly debugging repeat.

Ah, great. Would you mind plugging your AMI ID into the [ami_id param](https://github.com/hashicorp/terraform-aws-vault/blob/master/variables.tf#L14) of the root module, running `terraform apply`, and checking the cluster still boots up OK? You can use the [helper script](https://github.com/hashicorp/terraform-aws-vault/tree/master/examples/vault-examples-helper) to see if the cluster comes up properly.

> Issue in this case is the base image has a more restrictive umask causing files and folders to be created with a more restrictive policy.  This umask setting is a CIS hardening recommendation for securing servers

Ah, understood, thanks!

jasonmcintosh:
What automated tests do you have btw for this kinda stuff?  Been looking at smoke/integration style testing for terraform modules and not found a whole lot I like yet.

brikis98:
You can find the tests here: https://github.com/hashicorp/terraform-aws-vault/tree/master/test. We write them in Go, using a library we wrote called Terratest that has some shared utilities (e.g., for running `terraform apply`, `terraform destroy`, and lots of other useful stuff). We are still working on getting that library open source!

jasonmcintosh:
Worked beautifully with my AMI.  :)  

Note, one thing had to do (temporarily) was change subnet search in the data block for aws_subnet_ids - if you use any of the standard VPC creation (e.g. public/private subnet layouts), the default module doesn't work without changing the data lookup to search by tag.  

I can add another RFE issue for this.

brikis98:
Fantastic, thank you for testing!

brikis98:
https://github.com/hashicorp/terraform-aws-vault/releases/tag/v0.1.1

jasonmcintosh:
NP - sorry for the delay was sick most of the weekend.

Issue #40, opened 2018-02-06
Make the S3 backend optional


patoarvizu:
There has been some discussion around bringing back the option to configure S3 as a backend, on top of the Consul HA backend.

I'm opening this issue in preparation for a PR I'll be submitting shortly.

brikis98:
Fixed by #41. Thanks @patoarvizu!

PR #41, opened 2018-02-06
Add logic to make the s3 backend optional

patoarvizu:
I'm submitting this PR in connection with issue #40, although I'd like to get some assistance around two things:

- Since Terratest is not open sourced yet, I believe you have to run tests on your side, and
- Updating /_docs/architecture.png to reflect this change, or give the OK to roll back to the previous architecture diagram :)

I've updated the documentation as requested and I also added a separate example deployment, as well as made sure that the existing example deployments still worked.

I'm looking forward to your feedback!


modules/run-vault/run-vault
```diff
@@ -32,10 +32,17 @@ function print_usage {
   echo -e "  --log-level\t\tThe log verbosity to use with Vault. Optional. Default is $DEFAULT_LOG_LEVEL."
   echo -e "  --user\t\tThe user to run Vault as. Optional. Default is to use the owner of --config-dir."
   echo -e "  --skip-vault-config\tIf this flag is set, don't generate a Vault configuration file. Optional. Default is false."
+  echo -e "  --enable-s3-backend\tIf this flag is set, vault will generate the default.hcl configuration file with an S3 backend in addition to Consul."
```
brikis98: Update description here too.

modules/run-vault/run-vault
```diff
@@ -103,14 +110,33 @@ function generate_vault_config {
...
+  local readonly s3_bucket_region="$9"
   local readonly config_path="$config_dir/$VAULT_CONFIG_FILE"
 
   local instance_ip_address
   instance_ip_address=$(get_instance_ip_address)
 
   log_info "Creating default Vault config file in $config_path"
-  cat > "$config_path" <<EOF
-storage "consul" {
+  cat >> "$config_path" <<EOF
```
brikis98: It's a bit hard to visualize this config file when it's broken up across several steps and if/else statement. Also, the Consul storage config ends up being copy/pasted twice. 

Perhaps a cleaner approach is to:

1. Define an `s3_config` variable that is set to empty string by default, but if `enable_s3_backend` is `true`, you set it to the `storage "s3" { ... }` block.

1. Define a `consul_storage_type` variable that is set to `storage` by default, but gets reset to `ha_storage` if `enable_s3_backend` is `true`. You can then define the Consul configuration exactly once in the format: `$consul_storage_type "consul" { ... }`.

modules/run-vault/run-vault
```diff
@@ -228,6 +265,19 @@ function run {
...
+      --s3-bucket)
+        assert_not_empty "$key" "$2"
+        s3_bucket="$2"
+        shift
+        ;;
+      --s3-bucket-region)
+        assert_not_empty "$key" "$2"
+        s3_bucket_region="$2"
+        shift
+        ;;
```
brikis98: Around line 245 there are some `assert_not_empty` checks. If `enable_s3_backend` is `true`, you should add some `assert_not_empty` checks for the s3 bucket name & region configs.

modules/vault-cluster/main.tf
```diff
@@ -178,3 +178,33 @@ data "aws_iam_policy_document" "instance_role" {
...
+  tags {
+    Description = "Used for secret storage with Vault. DO NOT DELETE this Bucket unless you know what you are doing."
+  }
+}
+
+resource "aws_iam_role_policy" "vault_s3" {
+  count  = "${var.enable_s3_backend ? 1 : 0}"
+  name   = "vault_s3"
+  role   = "${aws_iam_role.instance_role.id}"
+  policy = "${data.aws_iam_policy_document.vault_s3.json}"
```
brikis98: Since this `aws_iam_policy_document` uses `count`, the recommended syntax in Terraform 0.11 is: `element(concat(data.aws_iam_policy_document.vault_s3.*.json, list("")), 0)`.

examples/vault-s3-backend/README.md
```diff
@@ -0,0 +1,48 @@
...
+[consul-cluster module](https://github.com/hashicorp/terraform-aws-consul/tree/master/modules/consul-cluster) from the Consul AWS Module.
+
+This example creates a Vault cluster spread across the subnets in the default VPC of the AWS account. Each of the
+servers in this example has [Dnsmasq](http://www.thekelleys.org.uk/dnsmasq/doc.html) installed (via the 
+[install-dnsmasq module](https://github.com/hashicorp/terraform-aws-consul/tree/master/modules/install-dnsmasq)) 
+which allows it to use the Consul server cluster for service discovery and thereby access Vault via DNS using the 
+domain name `vault.service.consul`. For an example of a Vault cluster
+that is publicly accessible, see [vault-cluster-public](https://github.com/hashicorp/terraform-aws-vault/tree/master/examples/vault-cluster-public).
+
+![Vault architecture](https://github.com/hashicorp/terraform-aws-vault/blob/master/_docs/architecture.png?raw=true)
```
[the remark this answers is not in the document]
patoarvizu: Just pushed some changes, but one note here is that I'm linking  to the eventual path once this is (hopefully) merged, so you don't have to re-link after the merge. I hope that's ok.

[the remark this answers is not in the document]
brikis98: Yea, that's a good idea, thx. Ideally, we'd use relative paths and avoid that problem, but the Terraform Registry requires absolute URLs :(

modules/run-vault/run-vault
```diff
@@ -103,14 +110,39 @@ function generate_vault_config {
...
+  local readonly s3_bucket_region="$9"
   local readonly config_path="$config_dir/$VAULT_CONFIG_FILE"
 
   local instance_ip_address
   instance_ip_address=$(get_instance_ip_address)
 
   log_info "Creating default Vault config file in $config_path"
-  cat > "$config_path" <<EOF
-storage "consul" {
+  listener_config=$(cat <<EOF
```
brikis98: We follow the [Google Bash Style Guide](https://google.github.io/styleguide/shell.xml), so please declare the new variables in this PR using `local` and, where possible, `readonly`. 

E.g.,:

```bash
local readonly consul_storage_type="storage"
local s3_config=""

if [[ "$enable_s3_backend" == "true" ]]; then
  ...
fi

local readonly listener_config=$(cat <<EOF
...
EOF
)

local readonly consul_storage_config=$(cat <<EOF
...
EOF
)
```

brikis98:
OK, I made a few minor fixes, and the tests are passing. Thanks!

brikis98:
https://github.com/hashicorp/terraform-aws-vault/releases/tag/v0.2.0

Issue #19, opened 2017-10-19
module vault-elb is not idempotent


kostyrev:
I'm creating ELB for vault with this:
```
module "vault_elb" {
  source                      = "git::git@github.com:hashicorp/terraform-aws-vault.git//modules/vault-elb"
  internal                    = true
  name                        = "${var.country}-${var.stage}-vault-elb"
  vpc_id                      = "${module.network-getter.vpc_id}"
  subnet_ids                  = "${module.network-getter.private_subnet_ids}"
  allowed_inbound_cidr_blocks = ["${data.aws_vpc.main.cidr_block}"]
}
```
everything applies with no problem. Thank you guys!
But after invoking `terraform plan` once again I see the following:
```
Terraform will perform the following actions:
  ~ module.vault_elb.aws_elb.vault
      availability_zones.#:          "3" => "0"
      availability_zones.1305112097: "us-east-1b" => ""
      availability_zones.3569565595: "us-east-1a" => ""
      availability_zones.986537655:  "us-east-1c" => ""


Plan: 0 to add, 1 to change, 0 to destroy.
```
```
$ terraform version
Terraform v0.10.7
```

brikis98:
99% of the time, these sorts of issues are bugs in Terraform itself. The vault-elb module is setting `availability_zones` to an empty list by default; my guess is that after creating the ELB, AWS fills in the AZ data automatically, so next time you run `plan`, Terraform figures that the empty AZ list means "remove all AZs". 

Perhaps file a bug in the [AWS provider](https://github.com/terraform-providers/terraform-provider-aws)?

kchugalinskiy:
```
data "aws_subnet" "private" {
  id = "${var.private_subnets}"

  count = "${length(var.private_subnets)}"
}

module "vault_elb" {
....blah-blah-blah....
  availability_zones          = "${data.aws_subnet.private.*.availability_zone}"
}
```
@kostyrev this may be a workaround for your problem :-)

@brikis98 maybe this should be used as a default value for `availability_zones` argument. This datasource is available since version 0.9.5 which is minimal in your case. Proof: https://github.com/hashicorp/terraform/blob/v0.9.5/website/source/docs/providers/aws/d/subnet.html.markdown

Guys, what do you think?

brikis98:
@kchugalinskiy Nice workaround. I'd welcome a PR for that!

kchugalinskiy:
[...]

PR #44, opened 2018-02-18
Remove the 'availability_zones' parameter

patoarvizu:
This is alternate approach **A** for #19, simply removing the `availability_zones` parameter. The main effect of this is that ELBs on EC2-classic will no longer be supported.

### DIFF
--- a/modules/vault-elb/main.tf
+++ b/modules/vault-elb/main.tf
@@ -20,7 +20,6 @@ resource "aws_elb" "vault" {
   connection_draining_timeout = "${var.connection_draining_timeout}"
 
   security_groups    = ["${aws_security_group.vault.id}"]
-  availability_zones = ["${var.availability_zones}"]
   subnets            = ["${var.subnet_ids}"]
 
   # Run the ELB in TCP passthrough mode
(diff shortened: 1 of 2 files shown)


brikis98:
Thanks for the PR! Code change looks good. How did you test it?

patoarvizu:
I deployed a cluster from the main example, then tried a straight up `terraform apply` right after, and it didn't flag any changes, then changed the `subnets` parameter and it correctly applied those changes, and setting them back would also cleanly bring it back to its original state.

I couldn't find anything in the `tests/` directory around the ELB, so I didn't add or modify anything there, but please let me know if I should. It is a bit hard to write new tests without Terratest though :)

brikis98:
> I couldn't find anything in the tests/ directory around the ELB, so I didn't add or modify anything there, but please let me know if I should. It is a bit hard to write new tests without Terratest though :)

Yea, sorry about that. We'll get it open sourced, I promise :)

In the meantime, we depend on helpful contributors to test things out manually. Thanks for the PR! Merging now.

brikis98:
https://github.com/hashicorp/terraform-aws-vault/releases/tag/v0.3.0

PR #45, opened 2018-02-18
Add lifecycle rule to ignore modifications to availability zones

patoarvizu:
This is alternate approach **B** for #19, adding a `lifecycle` rule to ignore changes to the `availability_zones` parameter. With this approach, ELBs on EC2-classic would still be supported, but changes to their availability zones won't be possible.


brikis98:
Thanks for the PR. I think having a feature that partially works would be more confusing than not having the feature at all, so I'll close this PR in favor of approach A in #19.

Issue #46, opened 2018-03-05
Option to create individual EC2 instances as Vault cluster rather than an ASG


jgiles:
In many ways creating individual EC2 instances is more desirable than an
Auto Scaling Group for a Vault cluster:

- Since instances typically require manual action to unseal, automatic
  creation of new instances doesn't buy you much.
- Individual instances are created with known IPs, making it easier to
  route requests to specific instances (e.g. for unsealing).
- Known IPs can be used to create DNS records referring to individual
  instances by name. This is as nice human affordance, but also greatly
  simplifies TLS communication with individual instances (which can hold
  a wildcard TLS cert valid for all instance DNS names).
- DNS discovery of instance addresses for unsealing avoids a
  chicken-and-egg problem with using the AWS CLI to look up instance
  addresses when you are using Vault to issue all your AWS CLI creds.

We've forked the repo to do this internally. The main difficulty is that you have to be a little more careful when updating your Vault AMI; destroying all the instances simultaneously leads to an outage until you get one unsealed. We use the `-target` option to do a two-phase update.

See conversation at https://github.com/hashicorp/vault/issues/764#issuecomment-173629647 for a little bit of context on ASG vs individual instances.

brikis98:
You make some pretty good points. However, a few thoughts in favor of the ASG:

* Vault Enterprise supports [auto-unseal](https://www.hashicorp.com/blog/vault-0-9#cloud-unseal), so auto recovery may be quite useful.
* Relying on the IPs of EC2 Instances isn't a good idea, as the IP will change if the instance crashes and is replaced. You can attach an ENI and rely on that as your static IP, but that's doable with an ASG as well by running a script on each Instance to attach the ENI during boot.
* You can attach DNS records to instances in an ASG as well by running a script on each Instance to make the Route 53 changes.
* One of the advantages of an ASG is that it has integration with an ELB. Combined with auto-unseal, this would actually make it possible to do a zero-downtime, rolling update that replaces one Vault node at a time, waits for it to pass health checks, and then repeats the process with the next one. We have code to do this for some other distributed systems (e.g., kafka) and hope to open source it in the future.

That said, I can see arguments in either direction. 

@KFishner Is there someone from the Vault team who could chime in with thoughts?

KFishner:
@amanoske what are your thoughts?

jgiles:
[...]

PR #47, opened 2018-03-05
Add option creating individual instances, not ASG.

jgiles:
Also add a module for setting up DNS records for a cluster of individual
instances, optionally includinge a SRV record resolving to all hosts.

In many ways creating individual EC2 instances is more desirable than an
Auto Scaling Group for a Vault cluster:

- Since instances typically require manual action to unseal, automatic
  creation of new instances doesn't buy you much.
- Individual instances are created with known IPs, making it easier to
  route requests to specific instances (e.g. for unsealing).
- Known IPs can be used to create DNS records referring to individual
  instances by name. This is as nice human affordance, but also greatly
  simplifies TLS communication with individual instances (which can hold
  a wildcard TLS cert valid for all instance DNS names).
- DNS discovery of instance addresses for unsealing avoids a
  chicken-and-egg problem with using the AWS CLI to look up instance
  addresses when you are using Vault to issue all your AWS CLI creds.

### DIFF
--- a/modules/vault-cluster/variables.tf
+++ b/modules/vault-cluster/variables.tf
@@ -42,6 +42,11 @@ variable "cluster_size" {
 # These parameters have reasonable defaults.
 # ---------------------------------------------------------------------------------------------------------------------
 
+variable "use_asg" {
+  description = "Use an EC2 Auto Scaling Group for the Vault cluster. If set to false, individual instances are created."
+  default = true
+}
+
 variable "subnet_ids" {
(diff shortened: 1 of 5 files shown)


jgiles:
This is just our fix for https://github.com/hashicorp/terraform-aws-vault/issues/46

I don't expect to merge it as-is, but wanted to put the code up. I'm happy to update docs, etc if the approach seems reasonable.

modules/vault-cluster/variables.tf
```diff
@@ -91,6 +96,12 @@ variable "cluster_extra_tags" {
   default = []
 }
 
+variable "instance_extra_tags" {
```
jgiles: Maybe there's a way to derive this from `cluster_extra_tags`, but I don't have the terraform-foo for it.

jgiles:
Abandoned, see https://github.com/hashicorp/terraform-aws-vault/issues/46#issuecomment-373743977

PR #55, opened 2018-04-09
Add name tag to security group

bruno:
To keep it consistent with the way it’s done in `terraform-aws-consul`


brikis98:
> Is there a test I should be submitting together with this PR @brikis98 ?

There are tests in the `test` folder, but we haven't yet open sourced the library we use with them, so until we do, you won't be able to. Sorry about that!

Merging now, thx for the PR.

brikis98:
https://github.com/hashicorp/terraform-aws-vault/releases/tag/v0.5.1

PR #56, opened 2018-04-09
Permit consul traffic in vault sg

bruno:
Adding the `security_group_rules` from `terraform-aws-consul` to permit the consul agents on the vault cluster to talk freely.

I used the `v0.3.1` tag (which is different from the one used in the example) as that's what I used in my local provisioning.

### DIFF
--- a/examples/vault-cluster-private/main.tf
+++ b/examples/vault-cluster-private/main.tf
@@ -70,12 +70,12 @@ data "template_file" "user_data_vault_cluster" {
 
 # ---------------------------------------------------------------------------------------------------------------------
 # PERMIT CONSUL SPECIFIC TRAFFIC IN VAULT CLUSTER
-# To allow our Vault servers consul agents to communication with consul agents and servers, we open up the consul
-# specific protocols and ports for consul traffic
+# To allow our Vault servers consul agents to communicate with other consul agents and participate in the LAN gossip,
+# we open up the consul specific protocols and ports for consul traffic
 # ---------------------------------------------------------------------------------------------------------------------
 
 module "security_group_rules" {
-  source = "github.com/hashicorp/terraform-aws-consul.git//modules/consul-security-group-rules?ref=v0.3.1"
+  source = "github.com/hashicorp/terraform-aws-consul.git//modules/consul-client-security-group-rules?ref=v0.3.1"
 
   security_group_id           = "${module.vault_cluster.security_group_id}"
 
--- a/examples/vault-cluster-private/main.tf
+++ b/examples/vault-cluster-private/main.tf
@@ -69,6 +69,19 @@ data "template_file" "user_data_vault_cluster" {
 }
 
 # ---------------------------------------------------------------------------------------------------------------------
+# PERMIT CONSUL SPECIFIC TRAFFIC IN VAULT CLUSTER
+# To allow our Vault servers consul agents to communication with consul agents and servers, we open up the consul
+# specific protocols and ports for consul traffic
+# ---------------------------------------------------------------------------------------------------------------------
+
+module "security_group_rules" {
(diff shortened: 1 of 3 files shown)


bruno:
@brikis98 this was tested in the same fashion as #55 by provisioning local dev infrastructure.

examples/vault-cluster-private/main.tf
```diff
@@ -68,6 +68,23 @@ data "template_file" "user_data_vault_cluster" {
   }
 }
 
+# ---------------------------------------------------------------------------------------------------------------------
+# PERMIT CONSUL SPECIFIC TRAFFIC IN VAULT CLUSTER
+# To allow our Vault servers consul agents to communication with consul agents and servers, we open up the consul
```
brikis98: s/communication/communicate/

examples/vault-cluster-private/main.tf
```diff
@@ -68,6 +68,23 @@ data "template_file" "user_data_vault_cluster" {
...
+# specific protocols and ports for consul traffic
+# ---------------------------------------------------------------------------------------------------------------------
+
+module "security_group_rules" {
+  source = "github.com/hashicorp/terraform-aws-consul.git//modules/consul-security-group-rules?ref=v0.3.1"
```
brikis98: Hm, I'm worried this might be overkill, and a bit of a security risk. When Consul is running in client mode, as it is on the Vault nodes, it should only listen on localhost to most ports. The only port that should be exposed to the outside world is the LAN gossip port. See [Consul Architecture](https://www.consul.io/docs/internals/architecture.html) for details.

That makes me think we should have a new, very simple module in `terraform-aws-consul` called `consul-client-security-group-rules` that opens up just this one port...

bruno: You're absolutely right, I just took the simplest route however opening up all that traffic _on the vault instances_ is not a good idea! I will modify this and submit that small module change on the `terraform-aws-consul` repo.

bruno:
Thanks for the feedback @brikis98 ... I gave your suggestion a try.

examples/vault-cluster-private/main.tf
```diff
@@ -68,6 +68,23 @@ data "template_file" "user_data_vault_cluster" {
...
 }
 
+# ---------------------------------------------------------------------------------------------------------------------
+# PERMIT CONSUL SPECIFIC TRAFFIC IN VAULT CLUSTER
+# To allow our Vault servers consul agents to communicate with other consul agents and participate in the LAN gossip,
+# we open up the consul specific protocols and ports for consul traffic
+# ---------------------------------------------------------------------------------------------------------------------
+
+module "security_group_rules" {
+  source = "github.com/hashicorp/terraform-aws-consul.git//modules/consul-client-security-group-rules?ref=v0.3.1"
```
brikis98: This version number will need to be bumped once your Consul PR is merged

bruno: @brikis98 thanks for merging the terraform-aws-consul PR ... do you think I should also update the S3 example to incorporate this change as well? And just a small guidance, should I also update the version number in the `consul-cluster` module?

brikis98: > do you think I should also update the S3 example to incorporate this change as well

If it's not too much of a pain, it would be great to update all the examples in this repo: the `main.tf` in the root, `examples/vault-cluster-private`, and `examples/vault-s3-backend`. 

> And just a small guidance, should I also update the version number in the consul-cluster module?

I think it'll be confusing to be using different terraform-aws-consul versions in different places, so probably a global search & replace to update to the latest version is the best option.

bruno:
Tested on my local vault dev cluster, provisioning it from scratch and updating a running one on a different AWS account.

brikis98:
https://github.com/hashicorp/terraform-aws-vault/releases/tag/v0.5.2

PR #61, opened 2018-04-17
Update api_addr and cluster_addr

briankassouf:
This addresses hashicorp/vault#4383

The api_addr key should not be using the cluster port. Also we have moved the api_addr and cluster_addr outside of the storage block. 

I was unable to test this change since I don't have an environment setup at the moment.

cc @boldandbusted

### DIFF
--- a/modules/run-vault/README.md
+++ b/modules/run-vault/README.md
@@ -91,6 +91,14 @@ available.
 ### Default configuration
 
 `run-vault` sets the following configuration values by default:
+* [api_addr](https://www.vaultproject.io/docs/configuration/index.html#api_addr): 
+      Set to `https://<PRIVATE_IP>:<PORT>` where `PRIVATE_IP` is the Instance's private IP fetched from
+      [Metadata](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html) and `PORT` is
+      the value passed to `--port`.  
+* [cluster_addr](https://www.vaultproject.io/docs/configuration/index.html#cluster_addr): 
+      Set to `https://<PRIVATE_IP>:<CLUSTER_PORT>` where `PRIVATE_IP` is the Instance's private IP fetched from
+      [Metadata](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html) and `CLUSTER_PORT` is
+      the value passed to `--cluster-port`.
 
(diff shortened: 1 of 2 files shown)


boldandbusted:
@briankassouf OK, wow, thank you for being so quick on this - but if you look at my config file in the reported issue, the address+port combination is *not* the same, unless I'm completely misreading the stanzas. :) I admit, it is late in the day for me.

briankassouf:
@boldandbusted excerpt of your config:
```
ha_storage "consul" {
  ...

  # HA settings
  cluster_addr  = "https://10.50.1.45:8201"
  redirect_addr = "https://10.50.1.45:8201"
}
```

brikis98:
Thanks for the PR!

> Also we have moved the api_addr and cluster_addr outside of the storage block.

Is this documented somewhere in the Vault release notes?

Also, we set a `cluster_addr` in the `listener "tcp" { ... }` block as well... Should that be using this new, top-level `cluster_addr` key?

> I was unable to test this change since I don't have an environment setup at the moment.

We can't accept PRs that aren't tested :-\

Any chance you'll be able to test this out and make sure the cluster still boots up, bootstraps, and accepts requests?

lawliet89:
I am having the same issue bootstrapping a new cluster with this module (due to https://github.com/hashicorp/vault/issues/4383).

@jefferai The documentation on Vault's [HA parameters](https://www.vaultproject.io/docs/configuration/index.html#high-availability-parameters) is very confusing: do these parameters go inside the `ha_storage` block or do they work as a top level parameter?

I have left them inside the `ha_storage` block and tweaked the port, and then Vault magically worked.

jefferai:
@brikis98 

> Also we have moved the api_addr and cluster_addr outside of the storage block.
> Is this documented somewhere in the Vault release notes?

It's still supported in there for backwards compat.

> Also, we set a cluster_addr in the listener "tcp" { ... } block as well... Should that be using this new, top-level cluster_addr key?

TCP block takes `cluster_address` not `cluster_addr`. That parallels `address` in the TCP block whereas the other parallels `api_addr`/(deprecated)`redirect_addr`, which should be at the top level but is supported in storage still.

brikis98:
@jefferai Did you test this PR before merging?

brikis98:
@jefferai Could you please let me know about whether you tested this code before merging, and if so, how?

PR #68, opened 2018-05-15
Value of 'count' cannot be computed fix

fsproru:
Hello, 

Ran into this issue with the recent version of terraform and vault_cluster:

```
module.pg_vault_cluster.module.security_group_rules.aws_security_group_rule.allow_api_inbound_from_security_group_ids: aws_security_group_rule.allow_api_inbound_from_security_group_ids: value of 'count' cannot be computed
```

Here is the fix I made for this issue. Let me know if you have any feedback.

### DIFF
--- a/modules/vault-cluster/variables.tf
+++ b/modules/vault-cluster/variables.tf
@@ -29,6 +29,10 @@ variable "allowed_inbound_security_group_ids" {
   type        = "list"
 }
 
+variable "allowed_inbound_security_group_count" {
+  description = "A count of allowed_inbound_security_group_ids used internally due to terraform limitation"
+}
+
 variable "user_data" {
   description = "A User Data script to execute while the server is booting. We recommend passing in a bash script that executes the run-vault script, which should have been installed in the AMI by the install-vault module."
 }
@@ -80,13 +84,13 @@ variable "cluster_extra_tags" {
   description = "A list of additional tags to add to each Instance in the ASG. Each element in the list must be a map with the keys key, value, and propagate_at_launch"
   type        = "list"
 
-  #example: 
+  #example:
   # default = [
   #   {
   #     key = "Environment"
   #     value = "Dev"
   #     propagate_at_launch = true
-  #   } 
+  #   }
   # ]
   default = []
 }
--- a/modules/vault-cluster/variables.tf
+++ b/modules/vault-cluster/variables.tf
@@ -30,7 +30,7 @@ variable "allowed_inbound_security_group_ids" {
 }
 
 variable "allowed_inbound_security_group_count" {
-  description = "A count of allowed_inbound_security_group_ids used internally due to terraform limitation"
+  description = "The number of entries in var.allowed_inbound_security_group_ids. Ideally, this value could be computed dynamically, but we pass this variable to a Terraform resource's 'count' property and Terraform requires that 'count' be computed with literals or data sources only."
 }
 
 variable "user_data" {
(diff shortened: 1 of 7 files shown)


modules/vault-cluster/variables.tf
```diff
@@ -29,6 +29,10 @@ variable "allowed_inbound_security_group_ids" {
   type        = "list"
 }
 
+variable "allowed_inbound_security_group_count" {
+  description = "A count of allowed_inbound_security_group_ids used internally due to terraform limitation"
```
josh-padnick: Let's add a more detailed explanation here so users don't get confused. How about something like:

_The number of entries in var.allowed_inbound_security_group_ids. Ideally, this value could be computed dynamically, but we pass this variable to a Terraform resource's 'count' property and Terraform requires that 'count' be computed with literals or data sources only._

modules/vault-security-group-rules/variables.tf
```diff
@@ -22,6 +22,10 @@ variable "allowed_inbound_security_group_ids" {
 # These parameters have reasonable defaults.
 # ---------------------------------------------------------------------------------------------------------------------
 
+variable "allowed_inbound_security_group_count" {
+  description = "A count of allowed_inbound_security_group_ids used internally due to terraform limitation"
+}
```
josh-padnick: See above comment.

josh-padnick:
You know, I just realized you'll also need to review each of the Terraform modules in the `/examples` folder, as well as the `main.tf` in the root of this repo and update them to use the new var.

Also, this is regrettably a backward-incompatible change in that we're adding a new required variable. That means users who didn't have an issue with this will now be forced to manually add this, but I'm not sure there's anything we can do to avoid that.

fsproru:
Hi @brikis98, let me answer your questions inline below:

> Could you let me know what your use case was that triggered this bug?

I'm using Terraform v0.11.7. The error comes up when I run `vault_cluster` module.

> You will need to update the example in the root main.tf and the other examples in the examples folder with this new param.

Updated.

> Could you run the tests? We can't run them automatically for external PRs for security reasons.

Sure thing, just ran it and they all pass ✅

```
vault-private-amzn 2018/05/18 12:43:31 module.vault_cluster.aws_iam_instance_profile.instance_profile: Destruction complete after 1s
vault-private-amzn 2018/05/18 12:43:31 module.vault_cluster.aws_iam_role.instance_role: Destroying... (ID: vault-test-k9O91w20180518193832367600000004)
vault-private-amzn 2018/05/18 12:43:31 module.vault_cluster.aws_iam_role.instance_role: Destructioncomplete after 1s
vault-private-amzn 2018/05/18 12:43:32 module.vault_cluster.aws_security_group.lc_security_group: Destruction complete after 3s
vault-private-amzn 2018/05/18 12:43:32
vault-private-amzn 2018/05/18 12:43:32 Destroy complete! Resources: 27 destroyed.
--- PASS: TestVaultClusterPrivateWithAmazonLinuxAmi (986.67s)
PASS
ok      terraform-aws-vault/test        1846.149s
```

@josh-padnick, great suggestion, I just updated the documentation per your request. Examples are also updated. You're exactly right, this is what we have to do due to the `count` limitation. Hopefully it will be less strict in the future and we can delete this code then.

Let me know if we should do any further changes. Looking forward to getting it merged so we can switch to the main branch in our project.

brikis98:
https://github.com/hashicorp/terraform-aws-vault/releases/tag/v0.7.0

PR #70, opened 2018-05-30
Fix `enable-s3-backend` documentation

Toady00:
The option in the script is `--enable-s3-backend`. But all other
references, in both the script and the README.md stated
`--s3-backend-enable` which wouldn't work.


brikis98:
https://github.com/hashicorp/terraform-aws-vault/releases/tag/v0.7.2

Issue #57, opened 2018-04-11
Vault UI support


holybit:
[Vault 0.10.0](https://github.com/hashicorp/vault/blob/master/CHANGELOG.md#0100-april-10th-2018) now has the UI as open source. Any changes planed to update vault-elb to support UI access?

brikis98:
Definitely a good thing to add. It's on our roadmap, but not in the short term, so in the meantime, PRs are very welcome!

PR #78, opened 2018-07-26
Added support for Vault UI

drmudgett:
Addresses Issue #57 
The module/run-vault/run-vault script now:
1 - Gathers the installed vault version number
2 - Compares that version to the minimum version required for UI support using a similar comparison mechanism as found in [this stack overflow response](https://stackoverflow.com/a/47897019/5627332).
3 - Adds "ui = true" if the installed version is >=0.10.0

This was tested using v.0.9.6 and v.0.10.4 and everything worked as expected.

### DIFF
--- a/modules/run-vault/run-vault
+++ b/modules/run-vault/run-vault
@@ -13,6 +13,7 @@ readonly EC2_INSTANCE_METADATA_URL="http://169.254.169.254/latest/meta-data"
 
 readonly SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 readonly SCRIPT_NAME="$(basename "$0")"
+readonly VAULT_VERSION=$(vault -v|awk '{print $2}'|tr -d v)
 
 function print_usage {
   echo
@@ -121,6 +122,12 @@ function generate_vault_config {
   instance_ip_address=$(get_instance_ip_address)
 
   log_info "Creating default Vault config file in $config_path"
+  local readonly ui_config=$(cat <<EOF
+ui = true
+
+EOF
+)
+
   local readonly listener_config=$(cat <<EOF
 listener "tcp" {
   address         = "0.0.0.0:$port"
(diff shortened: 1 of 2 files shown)


modules/run-vault/run-vault
```diff
@@ -159,6 +166,11 @@ api_addr      = "$api_addr"
 EOF
 )
 
+if [ $(echo "$VAULT_VERSION 0.10.0" | tr " " "\n" |sort --version-sort| head -n 1) = 0.10.0 ]; then
```
brikis98: Please extract this conditional into a named method (i.e., `vault_version_at_least`). Also, use `[[ ]]` instead of `[ ]`.

drmudgett: Just updated with the conditional in its own function.
I ran the automated tests, but I believe it failed because of a limitation on my AWS account.  Is there a more narrow test that I can run or a way to limit the resources that are created in other regions?

brikis98: Is the issue you can only use certain regions? If so, you can force Terratest to do so using the `TERRATEST_REGION ` environment variable: https://github.com/gruntwork-io/terratest/blob/master/modules/aws/region.go#L16

brikis98: You shouldn't need to tweak the timeout. The tests take ~15m to run. If they are timing out, it likely means something is failing. Deploy one of the examples, SSH to it, and check syslog for errors (e.g., during User Data).

drmudgett: Ok, I fixed some subnetting pieces on my side and the tests are passing now.

modules/run-vault/run-vault
```diff
@@ -104,6 +104,19 @@ function assert_is_installed {
   fi
 }
 
+function vault_version_at_least {
+  local readonly config_path=$1
+  local readonly ui_config=$2
+  local readonly VAULT_VERSION=$(/usr/local/bin/vault -v|awk '{print $2}'|tr -d v)
```
brikis98: I had to squint to figure out what this did. To help future maintainers, could you extract this into its own (one-line) named method (e.g., `get_vault_version`) and add a comment explaining what Vault's version output looks like and how this method strips the leading v.

modules/run-vault/run-vault
```diff
@@ -104,6 +104,19 @@ function assert_is_installed {
...
+
+  if [[ $(echo "$VAULT_VERSION 0.10.0" | tr " " "\n" |sort --version-sort| head -n 1) = 0.10.0 ]]
```
brikis98: I had to squint at this one even longer to understand that you were printing each version number on a line, sorting them, and making sure the earlier one is 0.10.0. Could you add a comment explaining this to make things easier for future maintainers?

drmudgett: Ok, comments have been added and the get_vault_version function has been separated out.  And everything is still passing the automated tests.

brikis98:
https://github.com/hashicorp/terraform-aws-vault/releases/tag/v0.9.0

PR #81, opened 2018-08-08
Add lb_tags to vault-elb module to provide additional tags

jcejohnson:
I am working on execution of the test suite but our accounts are locked down & preventing some of them from executing. I have, however, tested my change in our module's use of vault-elb.


brikis98:
https://github.com/hashicorp/terraform-aws-vault/releases/tag/v0.9.2

PR #83, opened 2018-08-10
Adding optional health_check_port variable to vault-elb

jcejohnson:
...so that vault-elb can, for instance, health check an 8201 LB against 8200

My cluster uses vault-elb to create a load-balancer for 8200 and a second LB for 8201. I want both health checks to hit 8200/sys/health. Thus the addition of this variable.
The new health_check_port defaults to zero which will cause vault_api_port to be used as before.

### DIFF
--- a/modules/vault-elb/main.tf
+++ b/modules/vault-elb/main.tf
@@ -31,7 +31,7 @@ resource "aws_elb" "vault" {
   }
 
   health_check {
-    target              = "${var.health_check_protocol}:${var.vault_api_port}${var.health_check_path}"
+    target              = "${var.health_check_protocol}:${var.health_check_port == 0 ? var.vault_api_port : var.health_check_port}${var.health_check_path}"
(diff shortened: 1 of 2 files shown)


jcejohnson:
Sadly, no. Our environment is very restricted in terms of regions, IAM, etc. I spent an afternoon hacking the test suite itself to try to work within our restrictions but, honestly, I wouldn't trust the results after those hacks.

I have tested it thoroughly in our codebase with the following configuration:

```
module "api_elb" {
  ...
  health_check_protocol = "HTTPS"
  health_check_path = "/v1/sys/health"
  ...
}

module "cluster_elb"  {
  ...
  vault_api_port = 8201
  health_check_protocol = "HTTPS"
  health_check_port = 8200
  health_check_path = "/v1/sys/health"
  ...
}
```

brikis98:
https://github.com/hashicorp/terraform-aws-vault/releases/tag/v0.9.3

PR #86, opened 2018-08-17
Adding tags to SG & S3 of vault-cluster and SG of vault-elb

jcejohnson:
Follow-up to PR#81

### DIFF
--- a/modules/vault-elb/main.tf
+++ b/modules/vault-elb/main.tf
@@ -58,6 +58,8 @@ resource "aws_security_group" "vault" {
   name        = "${var.name}-elb"
   description = "Security group for the ${var.name} ELB"
   vpc_id      = "${var.vpc_id}"
+
+  tags = "${var.sg_tags}"
 }
 
 resource "aws_security_group_rule" "allow_inbound_api" {
--- a/modules/vault-elb/variables.tf
+++ b/modules/vault-elb/variables.tf
@@ -117,6 +117,12 @@ variable "health_check_timeout" {
 }
 
 variable "lb_tags" {
-  description = "Tags to be applied to the load balancer."
+  description = "Tags to be applied to the ELB."
+  default     = {}
+}
+
+variable "sg_tags" {
+  description = "Tags to be applied to the ELB security group."
+  type        = "map"
   default     = {}
 }
--- a/modules/vault-cluster/main.tf
+++ b/modules/vault-cluster/main.tf
@@ -88,9 +88,7 @@ resource "aws_security_group" "lc_security_group" {
     create_before_destroy = true
   }
 
-  tags {
-    Name = "${var.cluster_name}"
-  }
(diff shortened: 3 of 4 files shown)


modules/vault-elb/variables.tf
```diff
@@ -117,6 +117,12 @@ variable "health_check_timeout" {
 }
 
 variable "lb_tags" {
-  description = "Tags to be applied to the load balancer."
+  description = "Tags to be applied to the ELB."
+  default     = {}
+}
+
+variable "sg_tags" {
```
brikis98: `security_group_tags`?

jcejohnson: Internal testing, same as https://github.com/hashicorp/terraform-aws-vault/pull/83. I have two clusters (blue/green) per region (three regions) in two accounts (non-prod & prod).

I went with 'sg_tags' to match https://github.com/hashicorp/terraform-aws-vault/pull/81/files where I introduced 'lb_tags'. In retrospect, I wish I had used 'load_balancer_tags' in #81 so that this one could be 'security_group_tags' to match consul. At the time I did #81 I had not yet looked into the consul side. It would be a breaking change but I would *really* like to go with the longer names if agree.

brikis98: Yea, longer names will be more readable. Breaking change is OK.

jcejohnson:
Updated. I'll run through our internal testing when I'm back on the VPN.

jcejohnson:
Ran through the usual local tests. Looks good.

modules/vault-elb/main.tf
```diff
@@ -38,7 +38,7 @@ resource "aws_elb" "vault" {
     timeout             = "${var.health_check_timeout}"
   }
 
-  tags = "${merge(map("Name", var.name), var.lb_tags)}"
+  tags = "${merge(var.load_balancer_tags, map("Name", var.name))}"
```
brikis98: Should `var.load_balancer_tags` override the default ones?

jcejohnson: What default ones?

If you mean "Name", no, I don't think so. Since 'Name' is typically used to set the name of the resource I think it would be wrong for var.load_balancer_tags to override Name with a value other than var.name (which is setting the resource name).

That's the same reasoning I followed on https://github.com/hashicorp/terraform-aws-consul/pull/74 with `tags = "${merge(map("Name", var.cluster_name), var.security_group_tags)}"`

jcejohnson:
Thanks! Will this become v0.9.4?

brikis98:
Sorry for the delay. The test failures in https://github.com/hashicorp/terraform-aws-vault/issues/88 are driving me batty and making it harder to verify commits.

New release is here:

https://github.com/hashicorp/terraform-aws-vault/releases/tag/v0.10.0

PR #89, opened 2018-08-23
Add support for circle ci 2.0

Etiene:
Adds support for circle 2.0


Etiene:
The intermittent failure is still happening! It is not unrelated to this PR though

PR #92, opened 2018-08-29
Add test for Vault Enterprise install

Etiene:
* Adds test for vault enterprise install

@infosecgithub

### DIFF
--- a/test/vault_cluster_enterprise_test.go
+++ b/test/vault_cluster_enterprise_test.go
@@ -1,0 +1,16 @@
+package test
+
+import (
+	"os"
+	"testing"
+)
+
+func TestVaultClusterEnterpriseWithUbuntuAmi(t *testing.T) {
+	t.Parallel()
+	runVaultEnterpriseClusterTest(t, "ubuntu16-ami", "ubuntu", os.Getenv("VAULT_AMI_TEMPLATE_VAR_DOWNLOAD_URL"))
+}
+
+func TestVaultClusterEnterpriseWithAmazonLinuxAmi(t *testing.T) {
+	t.Parallel()
+	runVaultEnterpriseClusterTest(t, "amazon-linux-ami", "ec2-user", os.Getenv("VAULT_AMI_TEMPLATE_VAR_DOWNLOAD_URL"))
+}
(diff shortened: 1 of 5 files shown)


Etiene:
Please review https://github.com/hashicorp/terraform-aws-vault/pull/87 before this one and I will rebase appropriately

test/vault_cluster_enterprise_test.go
```diff
@@ -0,0 +1,16 @@
...
+)
+
+func TestVaultClusterEnterpriseWithUbuntuAmi(t *testing.T) {
+	t.Parallel()
+	runVaultEnterpriseClusterTest(t, "ubuntu16-ami", "ubuntu", os.Getenv("VAULT_AMI_TEMPLATE_VAR_DOWNLOAD_URL"))
+}
+
+func TestVaultClusterEnterpriseWithAmazonLinuxAmi(t *testing.T) {
+	t.Parallel()
+	runVaultEnterpriseClusterTest(t, "amazon-linux-ami", "ec2-user", os.Getenv("VAULT_AMI_TEMPLATE_VAR_DOWNLOAD_URL"))
```
brikis98: The use of an env var here won't be obvious to future maintainers. Consider add a method that:

1. Explicitly checks if the env var is set
1. Reports an error that says "Env var xxx not found. Please set it to a Vault Enterprise download URL to run this test."
1. Add a comment to the method explaining that Vault Enterprise URLs should be kept secret, so we use an env var to avoid putting it directly in our code or logs.

test/vault_helpers.go
```diff
@@ -333,6 +333,81 @@ func runVaultWithS3BackendClusterTest(t *testing.T, packerBuildName string, sshU
...
 
+// Test the Vault enterprise cluster example by:
+//
+// 1. Copy the code in this repo to a temp folder so tests on the Terraform code can run in parallel without the
+//    state files overwriting each other.
+// 2. Build the AMI in the vault-consul-ami example with the given build name and the enterprise package
+// 3. Deploy that AMI using the example Terraform code
+// 4. SSH to a Vault node and initialize the Vault cluster
+// 5. SSH to each Vault node and unseal it
+// 5. SSH to a Vault node and make sure you can communicate with the nodes via Consul-managed DNS
```
brikis98: `5` is repeated twice (if you copy/pasted this comment block, might be a bug in the original!). Also, add another step here: check enterprise version of Vault got installed correctly.

PR #95, opened 2018-09-20
Example vault auth with EC2 metadata

Etiene:
- [x] example
 - [x] README
 - [x] test


examples/vault-ec2-auth/main.tf
```diff
@@ -0,0 +1,210 @@
...
+}
+
+resource "aws_instance" "example_auth_to_vault" {
+  ami           = "${var.ami_id}"
+  instance_type = "t2.micro"
+  subnet_id     = "${data.aws_subnet_ids.default.ids[0]}"
+  key_name      = "${var.ssh_key_name}"
+
+  # Uses the same security group as the vault cluster
+  # Plus an additional one that opens the port to our simple web server
```
brikis98: Why does it use the same security group as the vault cluster?

examples/vault-ec2-auth/main.tf
```diff
@@ -0,0 +1,210 @@
...
+  iam_instance_profile = "${aws_iam_instance_profile.example_instance_profile.name}"
+
+  tags {
+    Name = "${var.auth_server_name}"
+  }
+}
+
+# Using the same role as vault because it has consul_iam_policies_servers,
+# Which allows DescribeTags and enables this instance to connect to find and reach consul server
+# access the DNS registry for the vault server
```
brikis98: The permissions for this role come from this module, which you could also use directly: https://github.com/hashicorp/terraform-aws-consul/tree/master/modules/consul-iam-policies

examples/vault-ec2-auth/user-data-auth-client.sh
```diff
@@ -0,0 +1,99 @@
...
+# From: https://alestic.com/2010/12/ec2-user-data-output/
+exec > >(tee /var/log/user-data.log|logger -t user-data -s 2>/dev/console) 2>&1
+
+# These variables are passed in via Terraform template interpolation
+/opt/consul/bin/run-consul --client --cluster-tag-key "${consul_cluster_tag_key}" --cluster-tag-value "${consul_cluster_tag_value}"
+
+# Retrieves the pkcs7 certificate from instance metadata
+# The vault role name is filled by terraform
+readonly pkcs7=$(curl -s http://169.254.169.254/latest/dynamic/instance-identity/pkcs7 | tr -d '\n')
+readonly data=$(cat <<EOF
```
brikis98: Don't use `readonly` and call a subshell on the same line. Otherwise, the exit code from the subshell gets ignored, and you always get exit code 0 from the `readonly`, so the even if the command fails, the script keeps running.

examples/vault-ec2-auth/user-data-auth-client.sh
```diff
@@ -0,0 +1,99 @@
...
+{
+  "role": "${example_role_name}",
+  "pkcs7": "$pkcs7"
+}
+EOF
+)
+
+# run-consul is running on the background, so in case it fails we retry
+for i in $(seq 1 10); do login_output=$(curl --request POST --data "$data" "https://vault.service.consul:8200/v1/auth/aws/login") && s=0 && break || s=$? && sleep 20; done; (exit $s)
```
brikis98: I'm struggling to read this. Could you break it up into multiple lines? Also, the `s=0 && ...` logic is a bit hard to follow. I'd also recommend more retries with small `sleep` to speed things up.

See also this method which may be useful: https://github.com/hashicorp/terraform-aws-vault/blob/master/examples/vault-examples-helper/vault-examples-helper.sh#L123

examples/vault-ec2-auth/user-data-auth-client.sh
```diff
@@ -0,0 +1,99 @@
...
+# And use the token to perform operations on vault such as reading a secret
+readonly response=$(curl \
+  -H "X-Vault-Token: $token" \
+  -X GET \
+  https://vault.service.consul:8200/v1/secret/example_gruntwork)
+
+# If vault cli is installed we can also perform this operation by setting the necessary environment variables
+# export VAULT_TOKEN=$token
+# export VAULT_ADDR=https://vault.service.consul:8200
+# /opt/vault/bin/vault read secret/example_gruntwork
```
brikis98: 👍

examples/vault-ec2-auth/user-data-auth-client.sh
```diff
@@ -0,0 +1,99 @@
...
+
+# Serves the answer in a web server
+echo $response | jq -r .data.the_answer > index.html
+python -m SimpleHTTPServer 8080 &
```
brikis98: A couple thoughts:

1. I wouldn't recommend Python's `SimpleHTTPServer`, as it crashes/hangs with concurrent requests. With Ubuntu, we usually use busybox's built-in httpd instead: https://github.com/brikis98/terraform-up-and-running-code/blob/master/code/bash/03-terraform-state/user-data.sh#L9
1. I remember having to add `nohup` to prevent the server from exiting when the User Data script exits... But I'm not 100% sure that's necessary.

examples/vault-ec2-auth/user-data-vault.sh
```diff
@@ -0,0 +1,76 @@
...
+#
+# Vault does not store the generated master key. Without at least 3 key to
+# reconstruct the master key, Vault will remain permanently sealed!
+#
+# It is possible to generate new unseal keys, provided you have a quorum of
+# existing unseal keys shares. See "vault operator rekey" for more information.
+# ==========================================================================
+
+# Unseals the server with 3 keys from this output
+echo "$server_output" | head -n 3 | awk '{ print $4; }' | xargs -l /opt/vault/bin/vault operator unseal
```
brikis98: Faaaancy :)

The good news is that when you add auto unseal later, this stuff should, in theory, be simplified.

examples/vault-ec2-auth/main.tf
Etiene: I don't remember why I did this! Might have been related to thinking the consul agent would need it? But thinking about it now, I can't think why it would be necessary. I'll remove it

examples/vault-ec2-auth/main.tf
Etiene: Yes. I mentioned this in the doc I wrote. Gonna push it today

examples/vault-ec2-auth/user-data-auth-client.sh
Etiene: Oh, I didn't know that! Thanks :)

examples/vault-ec2-auth/user-data-vault.sh
Etiene: Following Josh's review, I'm adding a not about the auto unseal part :)

examples/vault-ec2-auth/user-data-auth-client.sh
Etiene: 1. I don't think this would be an issue here because, it is a simple example where the server is only there to expose one value and there wouldn't be concurrent requests coming in from the tests. Also this would work on Amazon Linux too. 

2. I considered adding nohup, but in the end it wasn't necessary

examples/vault-ec2-auth/user-data-auth-client.sh
Etiene: Ok, I will break it down

examples/vault-ec2-auth/user-data-auth-client.sh
brikis98: > I don't think this would be an issue here because, it is a simple example where the server is only there to expose one value and there wouldn't be concurrent requests coming in from the tests. Also this would work on Amazon Linux too.

All it takes is for one extra health check (e.g., from an ELB) or a single manual request while a test is running, and it'll crash... Not worth the potential for extra confusion IMO.

Etiene: Fair enough. But what about Amazon Linux then?

brikis98: There are lots of options: https://gist.github.com/willurd/5720255. I can't remember which we typically use for Amazon Linux. Perhaps Ruby is installed by default?

```
ruby -run -ehttpd . -p8000
```

Etiene: ruby isn't preinstalled in the ubuntu machine

Etiene:
Just seeing if there are any failures on tests now, but this is ready to re-review, now with docs & tests.

Etiene:
@infosecgithub  
@bgt101

examples/vault-ec2-auth/README.md
```diff
@@ -0,0 +1,171 @@
+# Vault authentication using EC2 metadata example
+
+This example shows how to use the metadata from an EC2 instance to authenticate
+to a [vault cluster][vault_cluster].
```
brikis98: This repo lives in the Terraform Registry, which has an unfortunate requirement: all links must be absolute.

examples/vault-ec2-auth/README.md
```diff
@@ -0,0 +1,171 @@
...
+1. Build a Vault and Consul AMI. See the [vault-consul-ami example][vault_consul_ami] documentation for
+   instructions. Make sure to note down the ID of the AMI.
+1. Install [Terraform](https://www.terraform.io/).
+1. Open `vars.tf`, set the environment variables specified at the top of the file, and fill in any other variables that
+   don't have a default. If you built a custom AMI, put the AMI ID into the `ami_id` variable. Otherwise, one of our
+   public example AMIs will be used by default. These AMIs are great for learning/experimenting, but are NOT
+   recommended for production use.
+1. Run `terraform init`.
+1. Run `terraform apply`.
+1. Run the [vault-examples-helper.sh script](https://github.com/hashicorp/terraform-aws-vault/tree/master/examples/vault-examples-helper/vault-examples-helper.sh) to
```
brikis98: Probably for this example, the user should curl the server to see that it was able to read the secret from vault, right?

examples/vault-ec2-auth/README.md
```diff
@@ -0,0 +1,171 @@
+# Vault authentication using EC2 metadata example
```
brikis98: Nice docs 👍

examples/vault-ec2-auth/main.tf
```diff
@@ -0,0 +1,210 @@
...
+}
+
+resource "aws_instance" "example_auth_to_vault" {
+  ami           = "${var.ami_id}"
+  instance_type = "t2.micro"
+  subnet_id     = "${data.aws_subnet_ids.default.ids[0]}"
+  key_name      = "${var.ssh_key_name}"
+
+  # Security group that opens the necessary ports for consul
+  # And simple security group that opens the port to our simple web server
```
brikis98: Simple twice in the same sentence.

examples/vault-ec2-auth/main.tf
```diff
@@ -0,0 +1,210 @@
...
+
+data "template_file" "user_data_vault_cluster" {
+  template = "${file("${path.module}/user-data-vault.sh")}"
+
+  vars {
+    aws_region               = "${data.aws_region.current.name}"
+    consul_cluster_tag_key   = "${var.consul_cluster_tag_key}"
+    consul_cluster_tag_value = "${var.consul_cluster_name}"
+    example_role_name        = "${var.example_role_name}"
+    example_secret           = "${var.example_secret}"
```
brikis98: Add comment explaining you'd never pass a secret this way. We are only doing it to make testing easier. In real usage, the secret should be stored by an operator calling Vault directly.

examples/vault-ec2-auth/user-data-auth-client.sh
```diff
@@ -0,0 +1,110 @@
+#!/bin/bash
+# This script is meant to be run in the User Data of each EC2 Instance while it's booting. The script uses the
+# run-consul script to configure and start Consul in client mode. Note that this script assumes it's running in an AMI
+# built from the Packer template in examples/vault-consul-ami/vault-consul.json.
```
brikis98: Update comment that this script is meant as an example of what Vault clients do to automatically auth.

examples/vault-ec2-auth/user-data-auth-client.sh
```diff
@@ -0,0 +1,110 @@
...
+
+# run-consul is running on the background, so we have to wait for it and
+# we also have to for wait for vault server to be booted and unsealed before it can accept this request
+# so in case this fails we retry.
+# The boolean operations with the exit status are there to temporarily circumvent the "set -e" at the
+# beginning of this script which exits the script immediatelly for error status.
+for i in $(seq 1 30); do
+  login_output=$(curl --request POST --data "$data" "https://vault.service.consul:8200/v1/auth/aws/login") && exit_status=0 || exit_status=$?
+  if [[ $exit_status -eq 0 ]]; then break; fi
+  sleep 10
```
brikis98: Add an `echo "Failed to auth to Vault. It may still be in the process of booting. Will sleep for 10 seconds and try again"` here.

examples/vault-ec2-auth/user-data-auth-client.sh
```diff
@@ -0,0 +1,110 @@
...
+EOF
+)
+
+# run-consul is running on the background, so we have to wait for it and
+# we also have to for wait for vault server to be booted and unsealed before it can accept this request
+# so in case this fails we retry.
+# The boolean operations with the exit status are there to temporarily circumvent the "set -e" at the
+# beginning of this script which exits the script immediatelly for error status.
+for i in $(seq 1 30); do
+  login_output=$(curl --request POST --data "$data" "https://vault.service.consul:8200/v1/auth/aws/login") && exit_status=0 || exit_status=$?
```
brikis98: Add an `echo "Attempting to authenticate to Vault..."`

examples/vault-ec2-auth/user-data-auth-client.sh
```diff
@@ -0,0 +1,110 @@
...
+  if [[ $exit_status -eq 0 ]]; then break; fi
+  sleep 10
+done; (exit $exit_status)
```
brikis98: This is hard to read, and it's not obvious to me why this wouldn't always exit the script. I think a better approach is to wrap this code in a function (e.g., `login_to_vault`) that calls `return` if login is successful or `exit $exit_status` if, after N retries, login was still unsuccessful.

examples/vault-ec2-auth/user-data-vault.sh
```diff
@@ -0,0 +1,87 @@
...
+# Initializes a vault server
+# run-vault is running on the background and we have to wait for it to be done,
+# so in case this fails we retry.
+# The boolean operations with the exit status are there to temporarily circumvent the "set -e" at the
+# beginning of this script which exits the script immediatelly for error status.
+for i in $(seq 1 20); do
+  server_output=$(/opt/vault/bin/vault operator init) && exit_status=0 || exit_status=$?
+  if [[ $exit_status -eq 0 ]]; then break; fi
+  sleep 5
+done; (exit $exit_status)
```
brikis98: Same comment here about adding comments (so it's clearer what the script is doing) and putting this in a function with `return` / `exit`.

examples/vault-ec2-auth/user-data-vault.sh
```diff
@@ -0,0 +1,87 @@
...
+#
+# Vault does not store the generated master key. Without at least 3 key to
+# reconstruct the master key, Vault will remain permanently sealed!
+#
+# It is possible to generate new unseal keys, provided you have a quorum of
+# existing unseal keys shares. See "vault operator rekey" for more information.
+# ==========================================================================
+
+# Unseals the server with 3 keys from this output
+# Please note that this is not how it should be done in production as it is not secure
```
brikis98: Not secure and also results in the root tokens being lost, so you'll never be able to unseal your Vault cluster again!

examples/vault-ec2-auth/user-data-vault.sh
```diff
@@ -0,0 +1,87 @@
...
+# other settings you can pick for EC2 metadata auth.
+# Read more at: https://www.vaultproject.io/api/auth/aws/index.html#create-role
+/opt/vault/bin/vault write \
+  auth/aws/role/${example_role_name}\
+  auth_type=ec2 \
+  policies=example-policy \
+  max_ttl=500h \
+  bound_ami_id=${ami_id}
+
+# Writes some secret, this secret is being written by terraform for test purposes
```
brikis98: Again, add a warning that you'd never pass the secret like this in plain text, and that we are only doing it so we can easily run automated tests in this example.

test/vault_helpers.go
```diff
@@ -409,6 +415,78 @@ func runVaultEnterpriseClusterTest(t *testing.T, packerBuildName string, sshUser
...
+//
+// 1. Copying the code in this repo to a temp folder so tests on the Terraform code can run in parallel without the
+//    state files overwriting each other.
+// 2. Building the AMI in the vault-consul-ami example with the given build name
+// 3. Deploying that AMI using the example Terraform code setting an example secret
+// 4. Waiting for Vault to boot, unseal and the client to fetch the secret
+// 5. Making a request to the webserver started by the auth client user data script to see if the secret was properly fetched
+func runVaultEC2AuthTest(t *testing.T, packerBuildName string) {
+	examplesDir := test_structure.CopyTerraformFolderToTemp(t, REPO_ROOT, VAULT_EC2_AUTH_PATH)
+	exampleSecret := "42"
```
brikis98: 👍

test/vault_helpers.go
```diff
@@ -435,6 +513,31 @@ func initializeAndUnsealVaultCluster(t *testing.T, asgNameOutputVar string, sshU
 	return cluster
 }
 
+func testRequestSecret(t *testing.T, terraformOptions *terraform.Options, expectedResponse string) {
+	instanceIP := terraform.Output(t, terraformOptions, OUTPUT_AUTH_CLIENT_IP)
+
+	curlCommand := shell.Command{
+		Command: "curl",
+		Args:    []string{"-s", fmt.Sprintf("%s:%s", instanceIP, "8080")},
+	}
```
brikis98: No need to do it this way. See https://github.com/gruntwork-io/terratest/blob/master/modules/http-helper/http_helper.go#L91 and other related methods.

examples/vault-ec2-auth/user-data-vault.sh
Etiene: Will add this in the comments

examples/vault-ec2-auth/main.tf
```diff
@@ -0,0 +1,210 @@
+# ---------------------------------------------------------------------------------------------------------------------
+# DEPLOY A VAULT SERVER CLUSTER AND A CONSUL SERVER CLUSTER IN AWS
+# This is an example of how to launch a vault cluster and then authenticate an instance to the cluster
+# ---------------------------------------------------------------------------------------------------------------------
+
+terraform {
+  required_version = ">= 0.11.7"
```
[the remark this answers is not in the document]
Etiene: Ok! I just put 0.11.7 as it is the one I was using to test things locally

examples/vault-ec2-auth/README.md
```diff
@@ -0,0 +1,171 @@
...
+EC2 auth is a process in which Vault relies on information about an EC2 instance
+trying to assume a desired authentication role. For different resources that are
+not EC2 instances, please refer to the `iam` auth method.
+
+The workflow is that the client trying to authenticate itself will send a
+signature in its login request, Vault verifies the signature with AWS, checks
+against a predefined authentication role, then returns a client token that the
+client can use for making future requests to vault. More details about the
+signature and how this works at the section [authenticating from an
+instance](#authenticating-from-an-instance)
```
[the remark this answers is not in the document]
Etiene: I'll just add these two paragraphs in there! 👍 Thanks

examples/vault-ec2-auth/README.md
```diff
@@ -0,0 +1,171 @@
...
+EOF
+)
+curl --request POST --data "$data" "https://vault.service.consul:8200/v1/auth/aws/login"
+```
+
+It is up to the client to decide how it handles the nonce. To read more about
+it, refer to the [Vault documentation on client nonce][nonce].
+
+To see the full script for authenticating check the [client user data
+script](user-data-auth-client.sh).
```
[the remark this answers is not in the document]
Etiene: According to the definition of nonce, yes you would be right. However vault does not gives us a new nonce on the second login attempt to be used on the third login attempt. (I just double checked this just to be sure!). All subsequent login attempts are made with the same nonce token received with the first login, making it not a "number used once".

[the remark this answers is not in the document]
josh-padnick: Oh, how interesting. I just checked the [Vault docs](https://www.vaultproject.io/docs/auth/aws.html#client-nonce) on the client nonce, and I now see that the "used once" concept is applied in the sense that only one unique principal (i.e. one unique EC2 Instance ID) can be use that nonce. The principal associated with that EC2 Instance ID can use the nonce in subsequent login attempt (assuming `disallow_reauthentication = true`), but a separate unique EC2 Instance ID could not use the nonce.

Given our new understanding, you might (a) link to the Vault docs on nonces, and (b) clarify the extent to which it's used once.

josh-padnick:
Anything left to do here before we're ready to merge?

examples/vault-ec2-auth/README.md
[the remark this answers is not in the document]
Etiene: (a) was already done. I'll do (b) now

Etiene:
It is now ready to merge :)

PR #96, opened 2018-10-02
Add vault iam auth example

Etiene:
This PR adds an example on how to login to Vault using the `iam` method. 
Improved docs and tests to come on a future PR


examples/vault-iam-auth/auth-signature-scripts/sign-request.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python
```
brikis98: Could you add a comment to this file explaining what it does and why it's in Python (as opposed to bash).

examples/vault-iam-auth/main.tf
```diff
@@ -0,0 +1,285 @@
...
+
+  tags {
+    Name = "${var.auth_server_name}"
+  }
+
+  # These two blocks below exist to provision a script that uses an AWS SDK to sign a request
+  provisioner "file" {
+    source      = "${path.module}/auth-signature-scripts"
+    destination = "/tmp"
+  }
```
brikis98: Curious why you use a provisioner here rather than packaging the python script into an AMI using Packer?

examples/vault-iam-auth/user-data-auth-client.sh
```diff
@@ -0,0 +1,104 @@
+#!/bin/bash
+# This script is meant to be run in the User Data of each EC2 Instance while it's booting. The script uses the
+# run-consul script to configure and start Consul in client mode. Note that this script assumes it's running in an AMI
+# built from the Packer template in examples/vault-consul-ami/vault-consul.json.
```
brikis98: This comment should probably be updated, as this script does quite a bit more now

examples/vault-iam-auth/user-data-auth-client.sh
```diff
@@ -0,0 +1,104 @@
...
+function retry {
+  local readonly cmd=$1
+  local readonly description=$2
+
+  for i in $(seq 1 30); do
+    echo "$description"
+
+    # The boolean operations with the exit status are there to temporarily circumvent the "set -e" at the
+    # beginning of this script which exits the script immediatelly for error status while not losing the exit status code
+    output=$(eval "$cmd") && exit_status=0 || exit_status=$?
```
brikis98: Are you using `output` anywhere? If not, any reason to use a subshell and store the output like that?

examples/vault-iam-auth/user-data-auth-client.sh
```diff
@@ -0,0 +1,104 @@
...
+    if [[ $exit_status -eq 0 ]]; then
+      return
+    fi
+    echo "$description failed. Will sleep for 10 seconds and try again."
+    sleep 10
+  done;
+
+  echo "$description failed."
```
brikis98: $description still failing after 10 retries

examples/vault-iam-auth/user-data-auth-client.sh
```diff
@@ -0,0 +1,104 @@
...
+  exit $exit_status
+}
+
+# Retrying this just in case terraform is still connecting and provisioning these files
```
brikis98: If you package the script via Packer, you can avoid this retry. Moreover, all the dependencies will already be installed, which is more idiomatic to how we build our modules.

examples/vault-iam-auth/user-data-auth-client.sh
```diff
@@ -0,0 +1,104 @@
...
+retry "chmod +x /tmp/auth-signature-scripts/install-dependencies.sh" "Chmod'ing signature script"
+/tmp/auth-signature-scripts/install-dependencies.sh
+
+# Creating a signed request to AWS STS API with header of server ID "vault.service.consul"
+# This request is to the method GetCallerIdentity of the AWS Security Token Service, which answers the question "who am I?"
+# This script uses python's AWS SDK boto3 to get necessary credentials and sign the request
```
brikis98: Add a note that we're using Python because signing requests in bash = pain

examples/vault-iam-auth/user-data-auth-client.sh
```diff
@@ -0,0 +1,104 @@
...
+# export AWS_ACCESS_KEY_ID=$(echo $creds | jq -r .AccessKeyId)
+# export AWS_SECRET_ACCESS_KEY=$(echo $creds | jq -r .SecretAccessKey)
+# export AWS_SESSION_TOKEN=$(echo $creds | jq -r .Token)
+
+
+# We can then use the client token from the login output once login was successful
+token=$(echo $output | jq -r .auth.client_token)
+
+# And use the token to perform operations on vault such as reading a secret
+response=$(curl \
```
brikis98: Any reason not to do the HTTP requests and most of the code above in python as well?

examples/vault-iam-auth/README.md
```diff
@@ -0,0 +1,109 @@
...
+}
+EOF
+
+vault write \
+  auth/aws/role/example-role
+  auth_type=ec2 \
+  policies=example-policy \
+  max_ttl=500h \
+  bound_iam_principal_arn=<ARN>
+```
```
josh-padnick: Show `iam` example here instead.

examples/vault-iam-auth/auth-signature-scripts/install-dependencies.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+function has_apt_get {
+  [[ -n "$(command -v apt-get)" ]]
+}
```
josh-padnick: I know this is a tiny bash script, but given that this script is targeted at Ubuntu, it may make sense to use the bash-commons functions we have for this: https://github.com/gruntwork-io/bash-commons/blob/master/modules/bash-commons/src/os.sh#L19-L24

examples/vault-iam-auth/main.tf
```diff
@@ -0,0 +1,285 @@
...
+  vars {
+    consul_cluster_tag_key   = "${var.consul_cluster_tag_key}"
+    consul_cluster_tag_value = "${var.consul_cluster_name}"
+    example_role_name        = "${var.example_role_name}"
+  }
+}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# ADDS A RULE TO OPEN PORT 8080 SINCE OUR EXAMPLE LAUNCHES A SIMPLE WEB SERVER
+# ---------------------------------------------------------------------------------------------------------------------
```
josh-padnick: Clarify that you're only doing this for automated tests and wouldn't normally do this in a prod environment.

examples/vault-iam-auth/main.tf
```diff
@@ -0,0 +1,285 @@
...
+
+# ---------------------------------------------------------------------------------------------------------------------
+# ADDS A POLICY TO THE VAULT CLUSTER ROLE SO VAULT CAN QUERY AWS IAM USERS AND ROLES
+# ---------------------------------------------------------------------------------------------------------------------
+
+resource "aws_iam_role_policy" "vault_iam" {
+  name   = "vault_iam"
+  role   = "${module.vault_cluster.iam_role_id}"
+  policy = "${data.aws_iam_policy_document.vault_iam.json}"
+}
```
josh-padnick: Does it make sense to create a dedicated module called `vault-aws-auth` which includes these permissions? Or perhaps one module for `vault-aws-auth-iam` and one for `vault-aws-auth-ec2` if the permissions are different? It seems like an essential aspect of setting up Vault when using the AWS auth method.

examples/vault-iam-auth/outputs.tf
```diff
@@ -0,0 +1,91 @@
...
+  value = "${module.vault_cluster.launch_config_name}"
+}
+
+output "iam_role_arn_vault_cluster" {
+  value = "${module.vault_cluster.iam_role_arn}"
+}
+
+output "iam_role_id_vault_cluster" {
+  value = "${module.vault_cluster.iam_role_id}"
+}
```
josh-padnick: There's some Gruntwork internal debate on this, but because `iam_role_id` is ambiguous on whether that refers to the IAM Role Name or ARN, I always prefer to explicitly name these `iam_role_name` or `iam_role_arn` to remove any ambiguity.

examples/vault-iam-auth/outputs.tf
```diff
@@ -0,0 +1,91 @@
...
+
+output "security_group_id_vault_cluster" {
+  value = "${module.vault_cluster.security_group_id}"
+}
```
josh-padnick: See above comment. `security_group_name` or `security_group_arn`.

examples/vault-iam-auth/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+When you create a Role in Vault, you define the Policies that are attached to that
+Role, how principals who assume that Role will re-authenticate, and for how long
+tokens issued for that role will be valid.
+
+In our example we create a simple Vault Policy that allows writing and reading from
+secrets in the path `secret` namespaced with the prefix `example_`, and then create
+a Vault Role that allows authentication from all instances with a specific `ami id`.
+You can read more about role creation and check which other instance metadata you can
+use on auth [here][create_role].
```
josh-padnick: Be sure to mention in the docs _when_ it makes sense to use IAM Auth vs. EC2 Auth.

examples/vault-iam-auth/auth-signature-scripts/sign-request.py
Etiene: Sure

examples/vault-iam-auth/main.tf
Etiene: I thought about that, but since this script is only used in this one example I decided to ship it like this. Should I ship it with packer even though it's for this one case?

examples/vault-iam-auth/user-data-auth-client.sh
Etiene: When retrying the login the output is used to get retrieve the client token

examples/vault-iam-auth/user-data-auth-client.sh
Etiene: If it's fine to do it even though it is used just for this example, sure!

examples/vault-iam-auth/user-data-auth-client.sh
Etiene: * Limiting scope of what happens behind the scenes. 
* Having the two examples of login (using vault cli or http request) below each other in the same file.

examples/vault-iam-auth/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+This example shows how to use the AWS IAM role attached to a resource to authenticate
+to a [vault cluster][vault_cluster].
+
+Vault provides multiple [auth methods][auth_methods] such as Username & Password, GitHub
+etc. Among those methods you will find AWS. The way it works is that it
+understands [AWS][aws_auth] as a trusted third party, and relies on AWS itself for affirming
+if an authentication source such as an EC2 Instance or other resources like a
+Lambda Function are legitimate sources or not. Basically, if AWS trusts the
+origin, then so do we.
```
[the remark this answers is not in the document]
Etiene: This PR was not intended to have docs. I just removed what was EC2 specific from the readme of the other example so it doesnt have a blank readme.

examples/vault-iam-auth/main.tf
brikis98: We should prob promote proper coding practices when we can. Of course, it's always a balance between ease of trying the example, security, and testability... But we already build AMIs for the Vault stuff, so we may as well add this script to that.

examples/vault-iam-auth/user-data-auth-client.sh
brikis98: I don't see any reference to `output` anywhere in this function? I see a reference further down in the code, outside of the function though... Is that intentional? Are you relying on the side effect of the `retry` function happening to set a variable called `output`? If so, that's quite confusing and non-obvious. I'd recommend restructuring to avoid that sort of leap.

examples/vault-iam-auth/main.tf
Etiene: Ok!

examples/vault-iam-auth/user-data-auth-client.sh
Etiene: Yes, I'm relying on the side-effect since bash functions only return integers. 
I could `echo` the output and then `$()` but then we loose the helpful echos with retry messages, so I opted for this compromise...  Do you know a better way where we can keep both?

brikis98: Yup! You can log to `stderr` and use `stdout` for return values. See https://github.com/gruntwork-io/bash-commons/blob/master/modules/bash-commons/src/log.sh.

Etiene: I could send a reference too... that might be the best compromise

Etiene: Ah! let me check that

examples/vault-iam-auth/README.md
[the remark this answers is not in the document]
Etiene: Also it is only signed on iam method, not on both methods

examples/vault-iam-auth/main.tf
Etiene: Why an independent module? If we are moving this out of here, it might as well just be an additional parameter for `vault-cluster`

examples/vault-iam-auth/outputs.tf
Etiene: That is not related to this PR, it would require refactoring `vault_cluster` and all other examples following your suggested pattern

Etiene: As per discussion on our zoom meeting, this would also be a breaking change.
For this reason we created an additional basecamp to-do categorised under a final project wrap-up to deal with this

examples/vault-iam-auth/main.tf
Etiene: As per discussion on our zoom meeting, we came to the conlusion it would be nice to have this as a separate module. But neither of us feel strongly about this. For this reason we created an additional basecamp to-do categorised under nice-to-have

Etiene:
Tests are failing after latest commit! I will investigate this, amd when they pass it should be ready to merge

examples/vault-iam-auth/user-data-auth-client.sh
Etiene: worked like a charm

Etiene:
Ok, most comments were addressed and the ones that weren't, the reason was that it could be addressed in a future PR, and now the CI issue has been fixed.
@brikis98 @josh-padnick do you want to give it another quick look before merge or should I go ahead? 

Notify:
@infosecgithub  
@bgt101

PR #97, opened 2018-10-04
Feature/launch template

jcejohnson:
The goal of this PR is to allow the module user to select either launch_configuration (traditional) or launch_template. Selecting launch_template should be backward compatible with the original implementation but with the ability to propagate tags to instance volumes. (With the caveat described below.)

As usual, I can only test this with our particular usecases. In that internal testing (one VPC per region, two vault clusters per VPC) I have validated that the resulting instances are identical to their launch_configuration counterparts and that the instance volumes have the expected tags.

I have not yet implemented a similar change to terraform-aws-consul. Doing so may result in changes to this PR if new edgecases are discovered there.

Caveat:
It only works when encrypted == "". This was reportedly addressed by AWS provider 1.34.0 but I'm still having issues. Ref: https://github.com/terraform-providers/terraform-provider-aws/issues/4553

### DIFF
--- a/modules/vault-cluster/main.tf
+++ b/modules/vault-cluster/main.tf
@@ -15,8 +15,10 @@ resource "aws_autoscaling_group" "autoscaling_group" {
 
   # launch_configuration = "${aws_launch_configuration.launch_configuration.name}"
 
+  depends_on = ["aws_iam_instance_profile.instance_profile", "aws_launch_template.launch_template"]
   launch_template {
-    id = "${aws_launch_template.launch_template.id}"
+    id      = "${aws_launch_template.launch_template.id}"
+    version = "$$Latest"
   }
   availability_zones  = ["${var.availability_zones}"]
   vpc_zone_identifier = ["${var.subnet_ids}"]
@@ -86,22 +88,26 @@ resource "aws_launch_template" "launch_template" {
   instance_type = "${var.instance_type}"
   user_data     = "${base64encode(var.user_data)}"
 
+  depends_on = ["aws_iam_instance_profile.instance_profile"]
+
   iam_instance_profile {
     name = "${aws_iam_instance_profile.instance_profile.name}"
   }
 
-  key_name               = "${var.ssh_key_name}"
-  vpc_security_group_ids = ["${concat(list(aws_security_group.lc_security_group.id), var.additional_security_group_ids)}"]
+  key_name = "${var.ssh_key_name}"
+
+  # Don't use vpc_security_group_ids when network_interfaces
+  # includes security_groups.
+  # vpc_security_group_ids = ["${concat(list(aws_security_group.lc_security_group.id), var.additional_security_group_ids)}"]
 
   placement {
     tenancy = "${var.tenancy}"
   }
-
-  # network_interfaces {
-  #   associate_public_ip_address = "${var.associate_public_ip_address}"
-  #   security_groups             = ["${concat(list(aws_security_group.lc_security_group.id), var.additional_security_group_ids)}"]
-  # }
-
+  network_interfaces {
+    associate_public_ip_address = "${var.associate_public_ip_address}"
+    delete_on_termination       = true
+    security_groups             = ["${concat(list(aws_security_group.lc_security_group.id), var.additional_security_group_ids)}"]
+  }
   ebs_optimized = "${var.root_volume_ebs_optimized}"
   block_device_mappings {
     device_name = "${data.aws_ami.ami.root_device_name}"
@@ -133,6 +139,23 @@ resource "aws_launch_template" "launch_template" {
   # https://terraform.io/docs/configuration/resources.html
   lifecycle {
     create_before_destroy = true
+  }
+}
+
+resource "null_resource" "update_launch_template" {
+  #
+  # Any time the launch configuration changes we need to use the CLI to create a
+  # new version of the template without the encryption method. I have only tested
+  # this where var.ebs_encryption is false.
+  # Got the idea & code from:
+  #   https://github.com/terraform-providers/terraform-provider-aws/issues/4553#issuecomment-414716171
+  #
+  triggers {
+    launch_template_version = "${aws_launch_template.launch_template.latest_version}"
+  }
+
+  provisioner "local-exec" {
+    command = "${var.ebs_encryption} || (sleep 5; aws ec2 create-launch-template-version --launch-template-id ${aws_launch_template.launch_template.id} --version-description turnOffEncryption --source-version '$$Latest' --launch-template-data '{\"BlockDeviceMappings\": [{\"DeviceName\": \"${data.aws_ami.ami.root_device_name}\", \"Ebs\": { \"VolumeSize\": ${var.root_volume_size}, \"VolumeType\": \"${var.root_volume_type}\" }}]}')"
   }
 }
 
@@ -252,8 +275,9 @@ resource "aws_s3_bucket" "vault_storage" {
   force_destroy = "${var.force_destroy_s3_bucket}"
 
   tags = "${merge(
-    map("Description", "Used for secret storage with Vault. DO NOT DELETE this Bucket unless you know what you are doing."),
-    var.s3_bucket_tags)
+    var.s3_bucket_tags,
+    map("Description", "Used for secret storage with Vault. DO NOT DELETE this Bucket unless you know what you are doing.")
+    )
   }"
 }
 
--- a/modules/vault-cluster/main.tf
+++ b/modules/vault-cluster/main.tf
@@ -11,25 +11,60 @@ terraform {
 # ---------------------------------------------------------------------------------------------------------------------
 
 resource "aws_autoscaling_group" "autoscaling_group" {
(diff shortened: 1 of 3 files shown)


modules/vault-cluster/variables.tf
```diff
@@ -197,3 +197,32 @@ variable "force_destroy_s3_bucket" {
   description = "If 'configure_s3_backend' is enabled and you set this to true, when you run terraform destroy, this tells Terraform to delete all the objects in the S3 bucket used for backend storage. You should NOT set this to true in production or you risk losing all your data! This property is only here so automated tests of this module can clean up after themselves. Only used if 'enable_s3_backend' is set to true."
   default     = false
 }
+
+# Launch Template Extensions
+
+variable "asg_launch_mechanism" {
+  description = "Select between launch_config-driven or launch_template-driven autoscaling group."
+  default     = "launch_config"
```
Etiene: On line 14 of `main.tf` the expected value is `launch_configuration`, not `launch_config`. 

It's also worth adding here more clearly what are the expected values, or even maybe doing this in a different way, such as using booleans, to avoid confusion?

Another thing is that I think this is lacking an explanation of why you would choose one way over another. If not here, maybe a small note on the Readme?

modules/vault-cluster/main.tf
```diff
@@ -72,6 +111,77 @@ resource "aws_launch_configuration" "launch_configuration" {
...
+    # Instance tags are already handled by the autoscaling group
+    resource_type = "volume"
+
+    tags = "${merge(
+      var.volume_extra_tags,
+      map("key", var.cluster_tag_key, "value", var.cluster_name)
+      )
+    }"
+  }
+  # Important note: whenever using a launch configuration with an auto scaling group, you must set
```
brikis98: Is this true with a launch template too?

modules/vault-cluster/main.tf
```diff
@@ -72,6 +111,77 @@ resource "aws_launch_configuration" "launch_configuration" {
...
+      encrypted             = "${var.root_volume_ebs_encryption}"
+      volume_type           = "${var.root_volume_type}"
+      volume_size           = "${var.root_volume_size}"
+      delete_on_termination = "${var.root_volume_delete_on_termination}"
+    }
+  }
+  tags = "${var.launch_template_tags}"
+  tag_specifications {
+    # Instance tags are already handled by the autoscaling group
+    resource_type = "volume"
```
brikis98: I must admit that this is a lot of extra code to maintain just to be able to tag EBS volumes. Is there no other way to do it?

modules/vault-cluster/main.tf
```diff
@@ -88,7 +198,7 @@ resource "aws_security_group" "lc_security_group" {
     create_before_destroy = true
   }
 
-  tags = "${merge(map("Name", var.cluster_name), var.security_group_tags)}"
+  tags = "${merge(var.security_group_tags, map("Name", var.cluster_name))}"
```
brikis98: Why reverse the order of the tags?

modules/vault-cluster/main.tf
```diff
@@ -29,10 +66,11 @@ resource "aws_autoscaling_group" "autoscaling_group" {
   wait_for_capacity_timeout = "${var.wait_for_capacity_timeout}"
 
   tags = ["${concat(
+    var.cluster_extra_tags,
```
brikis98: Why reverse the order of the tags?

modules/vault-cluster/outputs.tf
```diff
@@ -1,5 +1,7 @@
 output "asg_name" {
-  value = "${aws_autoscaling_group.autoscaling_group.name}"
+  # This is safe because asg_launch_mechanism will only allow one of aws_autoscaling_group.autoscaling_group.*
+  # or aws_autoscaling_group.lt_autoscaling_group.* to be non-empty.
+  value = "${join("",concat(aws_autoscaling_group.autoscaling_group.*.name,aws_autoscaling_group.lt_autoscaling_group.*.name))}"
```
brikis98: Our typical approach for this is: `${element(concat(x, y), 0)}`.

modules/vault-cluster/variables.tf
```diff
@@ -197,3 +197,37 @@ variable "force_destroy_s3_bucket" {
   description = "If 'configure_s3_backend' is enabled and you set this to true, when you run terraform destroy, this tells Terraform to delete all the objects in the S3 bucket used for backend storage. You should NOT set this to true in production or you risk losing all your data! This property is only here so automated tests of this module can clean up after themselves. Only used if 'enable_s3_backend' is set to true."
   default     = false
 }
+
+# Launch Template Extensions
+
+variable "asg_launch_mechanism" {
+  # Choices are launch_configuration and launch_template.
+  # launch_template gives you the opportunithy to declare volume encryption and
```
brikis98: You can enable volume encryption with a launch config too: https://www.terraform.io/docs/providers/aws/r/launch_configuration.html#encrypted

modules/vault-cluster/main.tf
jcejohnson: To be honest, I'm not entirely sure that it is (still?) true for launch_configuration.  My code is uses the vault-cluster default (true) and I haven't made any special configuration for the other resources.

modules/vault-cluster/main.tf
jcejohnson: I agree. My hope is that others will find other reasons why launch_template is a better alternative for their usecases and can expand on what I've started here.

The reason for the long delay between my initial submission & issue and this PR was due to the aws provider issue mentioned elsewhere. In the meantime, I was able to find a workaround via user-data script which inspects the instance & applies its tags to the instance volume. Of course, that requires that the instance be able to query its own tags and apply tags to its instance.

So, yes, there is another way to do it but it's a hack. And, from what I gather, launch templates are the new hot thing and presumably the preferred option going forward.

modules/vault-cluster/main.tf
jcejohnson: I don't want var.security_group_tags to inadvertently change the value of the Name tag.

modules/vault-cluster/main.tf
jcejohnson: I don't want var.cluster_extra_tagsto inadvertently change the value of the Name tag.

modules/vault-cluster/outputs.tf
jcejohnson: Ah. Nice. Fixed.

modules/vault-cluster/variables.tf
jcejohnson: Indeed it can. Comment updated.
(FWIW, that ability isn't currently exposed in this module)

jcejohnson:
So, the upshot of it is this...
- There's a lot of new code here to add the option to use launch_template instead of launch_configuration
- Some (i.e. - lt_autoscaling_group) of the new code is redundant because aws_autoscaling_group cannot be given both a launch_template and a launch_configuration even if one is empty
- Currently, the new vault-cluster features enabled by choosing launch_template are:
  - volume encryption (possible with launch_configuration but not exposed by this module)
  - specifying volume tags (not possible with launch_configuration but can be worked around via user-data)
- launch_template offers more features than launch_configuration, whether or not those are applicable to vault-cluster remains to be seen
- If this PR is accepted I will followup with a similar change to terraform-aws-consul to give it the ability to specify volume tags

modules/vault-cluster/main.tf
brikis98: Hm, I'm tempted to say that's a feature, not a bug, as it allows the user full control over the `Name` tag in case the default isn't what they want...

modules/vault-cluster/variables.tf
brikis98: We probably should expose the ability to encrypt EBS volumes in the module for both launch configs and launch templates...

modules/vault-cluster/main.tf
```diff
@@ -11,10 +11,47 @@ terraform {
...
+  tags = ["${concat(
+    var.cluster_extra_tags,
+    list(
+      map("key", var.cluster_tag_key, "value", var.cluster_name, "propagate_at_launch", true)
+      )
+    )
+  }"]
+}
+
+# An alternate autoscaling group that uses a launch_template
```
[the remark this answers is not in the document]
brikis98: Bummer

brikis98:
One thought for you: is it worth waiting until Terraform 0.12 comes out to add this feature? It will support both dynamic blocks and `null` defaults, so we'd be able to do this with just one resource rather than several.

jcejohnson:
> One thought for you: is it worth waiting until Terraform 0.12 comes out to add this feature? It will support both dynamic blocks and `null` defaults, so we'd be able to do this with just one resource rather than several.

I think that is an excellent idea!

brikis98:
OK, great. Hopefully, 0.12 will be out before too long, and we can make this PR simpler and get it merged. Thanks!

ozbillwang:
Any updates for this PR?

jcejohnson:
No. I've been re-tasked to work on GCP adoption and haven't been able to get back to this. If there is no other interest it's probably best to just close this.

PR #99, opened 2018-10-09
Add example of applying enterprise license on auto-unseal example

Etiene:
* Added example of applying enterprise license 
* Updated from variables for kms key to data source through key alias

### DIFF
--- a/examples/vault-auto-unseal/main.tf
+++ b/examples/vault-auto-unseal/main.tf
@@ -72,7 +72,6 @@ data "template_file" "user_data_vault_cluster" {
 
     kms_key_id                   = "${data.aws_kms_alias.vault-example.target_key_id}"
     aws_region                   = "${data.aws_region.current.name}"
-    vault_enterprise_license_key = "${var.vault_enterprise_license_key}"
   }
 }
 
--- a/examples/vault-auto-unseal/main.tf
+++ b/examples/vault-auto-unseal/main.tf
@@ -7,6 +7,10 @@ terraform {
   required_version = ">= 0.11.0"
 }
 
+data "aws_kms_alias" "vault-example" {
+  name = "alias/${var.auto_unseal_kms_key_alias}"
+}
+
 # ---------------------------------------------------------------------------------------------------------------------
 # DEPLOY THE VAULT SERVER CLUSTER
 # ---------------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


examples/vault-auto-unseal/main.tf
```diff
@@ -66,8 +70,9 @@ data "template_file" "user_data_vault_cluster" {
     consul_cluster_tag_key   = "${var.consul_cluster_tag_key}"
     consul_cluster_tag_value = "${var.consul_cluster_name}"
 
-    kms_key_id = "${var.auto_unseal_kms_key_id}"
-    aws_region = "${data.aws_region.current.name}"
+    kms_key_id                   = "${data.aws_kms_alias.vault-example.target_key_id}"
+    aws_region                   = "${data.aws_region.current.name}"
+    vault_enterprise_license_key = "${var.vault_enterprise_license_key}"
```
brikis98: Is the license key itself considered a secret? i.e., do we need to tell users to encrypt it for production use cases? Of course, this is a bit of a chicken-and-egg, as Vault itself is what we'd generally use to manage secrets...

examples/vault-auto-unseal/user-data-vault.sh
```diff
@@ -22,3 +55,32 @@ readonly VAULT_TLS_KEY_FILE="/opt/vault/tls/vault.key.pem"
...
   --auto-unseal-kms-key-region "${aws_region}"
+
+# Initializes the vault server
+# Retries this a number of times because run-vault is running on the background and
+# we need to wait for this to finish.
+# This userdata file is present in all instances in the cluster, but the code below
+# will only succeed in the first one that runs it, the others will fail with "the server has
+# already been initialized" and the script will be interrupted. This is not an
+# issue, and, for the purpose of this example, it is simpler than reorganizing the
+# code in such a way that the nodes receive different user data files.
```
brikis98: Hmmm... but what about production use cases? Should we tell users to run `init` manually against one node? Or perhaps better, should we detect the "already initialized" error and ignore it based on the assumption one of the other nodes already did the initialization? I'd rather avoid the script exiting with an error, as that will get very confusing to debug (e.g., if later on we add something after this line that we expect to run, but now doesn't on 2 out of 3 nodes).

examples/vault-auto-unseal/main.tf
Etiene: Following discussion with Josh we decided to treat it with moderate secrecy and not encrypt it

examples/vault-auto-unseal/user-data-vault.sh
Etiene: Following discussion with Josh, we decided to move this bit from the user data to the go tests since it should be manually done by the operator one time. This would then eliminate this issue. But I will leave it as a comment in the user data.

brikis98: Should it still be a manual operation even if you have auto-unseal and could therefore boot up a cluster completely automatically?

Etiene: That is a fair point. But then we have to make an informed decision to suggest on how to handle the root token in production, dont we?

brikis98: Oh, I totally forgot that's when you get a root token. Yea, never mind, best that's handled manually.

Etiene: ok 👍

PR #100, opened 2018-10-10
Test/iam auth

Etiene:
* Refactors vault test helper
* Bumps version of consul cluster module on examples
* Adds vault iam auth example

There still an intermittent test issue going on (from before this PR, which continues to happen through this PR). Josh suggested to use Eugene's new feature. I agree! I am thinking about tackling it in a different PR though. What do you think?

### DIFF
--- a/test/vault_helpers.go
+++ b/test/vault_helpers.go
@@ -108,7 +108,7 @@ func runVaultPrivateClusterTest(t *testing.T, packerBuildName string, sshUserNam
 	})
 
 	test_structure.RunTestStage(t, "deploy", func() {
-		deployCluster(t, examplesDir, random.UniqueId(), map[string]interface{}{})
+		deployCluster(t, examplesDir, random.UniqueId(), nil)
 	})
 
 	test_structure.RunTestStage(t, "validate", func() {
@@ -255,7 +255,7 @@ func runVaultEnterpriseClusterTest(t *testing.T, packerBuildName string, sshUser
 	})
 
 	test_structure.RunTestStage(t, "deploy", func() {
-		deployCluster(t, examplesDir, random.UniqueId(), map[string]interface{}{})
+		deployCluster(t, examplesDir, random.UniqueId(), nil)
 	})
 
 	test_structure.RunTestStage(t, "validate", func() {
@@ -275,8 +275,9 @@ func runVaultEnterpriseClusterTest(t *testing.T, packerBuildName string, sshUser
 //    state files overwriting each other.
 // 2. Building the AMI in the vault-consul-ami example with the given build name
 // 3. Deploying that AMI using the example Terraform code setting an example secret
-// 4. Waiting for Vault to boot, unseal and the client to fetch the secret
-// 5. Making a request to the webserver started by the auth client user data script to see if the secret was properly fetched
+// 4. Waiting for Vault to boot, then unsealing the server, creating a Vault Role to allow logins from instances with a specific EC2 property and writing the example secret
+// 5. Waiting for the client to login, read the secret and launch a simple web server with the contents read
+// 6. Making a request to the webserver started by the auth client
 func runVaultEC2AuthTest(t *testing.T, packerBuildName string) {
 	examplesDir := test_structure.CopyTerraformFolderToTemp(t, REPO_ROOT, VAULT_EC2_AUTH_PATH)
 	exampleSecret := "42"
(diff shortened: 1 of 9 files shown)


josh-padnick:
> I am thinking about tackling it in a different PR though. What do you think?

Sounds like a good idea. I'll review the rest of this PR now.

test/vault_helpers.go
```diff
@@ -379,29 +255,7 @@ func runVaultEnterpriseClusterTest(t *testing.T, packerBuildName string, sshUser
...
-				VAR_SSH_KEY_NAME:           keyPair.Name,
-			},
-			EnvVars: map[string]string{
-				ENV_VAR_AWS_REGION: awsRegion,
-			},
-		}
-		test_structure.SaveTerraformOptions(t, examplesDir, terraformOptions)
-
-		terraform.InitAndApply(t, terraformOptions)
+		deployCluster(t, examplesDir, random.UniqueId(), map[string]interface{}{})
```
josh-padnick: Can you replace ` map[string]interface{}{}` with just `nil`?

test/vault_helpers.go
```diff
@@ -428,57 +282,56 @@ func runVaultEC2AuthTest(t *testing.T, packerBuildName string) {
...
-		awsRegion := aws.GetRandomRegion(t, nil, nil)
-		test_structure.SaveString(t, examplesDir, SAVED_AWS_REGION, awsRegion)
+// Test the Vault IAM authentication example by:
+//
+// 1. Copying the code in this repo to a temp folder so tests on the Terraform code can run in parallel without the
+//    state files overwriting each other.
+// 2. Building the AMI in the vault-consul-ami example with the given build name
+// 3. Deploying that AMI using the example Terraform code setting an example secret
+// 4. Waiting for Vault to boot, unseal and the client to fetch the secret
+// 5. Making a request to the webserver started by the auth client user data script to see if the secret was properly fetched
```
josh-padnick: Nice description of the test, but it doesn't mention Vault IAM auth explicitly anywhere? Perhaps clarify in the last line that you're using Vault IAM Auth to auth the client? Also can you make it clear that you're not "waiting for vault" to unseal; rather, the test will unseal Vault.

test/vault_helpers.go
```diff
@@ -487,6 +340,60 @@ func runVaultEC2AuthTest(t *testing.T, packerBuildName string) {
...
+
+	for key, value := range map[string]interface{}{
+		VAR_AMI_ID:                 amiId,
+		VAR_VAULT_CLUSTER_NAME:     fmt.Sprintf("vault-test-%s", uniqueId),
+		VAR_CONSUL_CLUSTER_NAME:    fmt.Sprintf("consul-test-%s", uniqueId),
+		VAR_CONSUL_CLUSTER_TAG_KEY: fmt.Sprintf("consul-test-%s", uniqueId),
+		VAR_SSH_KEY_NAME:           keyPair.Name,
+	} {
+		terraformVars[key] = value
+	}
```
josh-padnick: This is a clever but somewhat unusual way to just merge two maps. Can you make this more clear by writing your own mini `merge` function that's more generic in nature and then call that? Should make the code more readable.

test/vault_helpers.go
```diff
@@ -487,6 +340,60 @@ func runVaultEC2AuthTest(t *testing.T, packerBuildName string) {
...
+
+	terraformOptions := &terraform.Options{
+		TerraformDir: examplesDir,
+		Vars:         terraformVars,
+		EnvVars: map[string]string{
+			ENV_VAR_AWS_REGION: awsRegion,
+		},
```
josh-padnick: Given that we're dynamically generating all the vars, should we just pass in the awsRegion directly to the Terraform vars and avoid the use of EnvVars altogether?

Etiene: We could. I was just following the previous examples that use `data "aws_region" "current" {}` and get it through env. That could be removed and we can pass it as vars, but I honestly don't see what's the difference and what would be the advantage of doing that over what was being done before. Could you help me understand?

test/vault_helpers.go
Etiene: sure!

test/vault_helpers.go
Etiene: Sure!

test/vault_helpers.go
Etiene: ok!

Etiene:
> These tests run with the open source Vault, not Vault Enterprise, correct?

Well, the vault enterprise cluster test runs with Vault Enterprise installed, so will the auto unseal test that I'm working right now (coming on another PR). All the others run with the open source vault.

test/vault_helpers.go
josh-padnick: > but I honestly don't see what's the difference

Actually, that was my thought exactly. But if there's no difference, and you see a different way of handling one particular var then it makes the reader question why it's being used that way (as I did). So I'm just trying to minimize the potential confusion readers will have.

josh-padnick:
> Well, the vault enterprise cluster test runs with Vault Enterprise installed, so will the auto unseal test that I'm working on right now (coming on another PR). All the others run with the open source vault.

So, this is Vault open source then? My point is that we should default to using Vault open source since it's the free version and only use Vault Enterprise in the examples when the examples specifically calls for it.

Etiene:
> So, this is Vault open source then? My point is that we should default to using Vault open source since it's the free version and only use Vault Enterprise in the examples when the examples specifically calls for it.

Yes, this is exactly how it works already

test/vault_helpers.go
Etiene: I see! I'll take a note to do this in a separate PR if you don't mind because I just did the vault auto-unseal branching off this one

Etiene:
Notify 
@infosecgithub  
@bgt101

PR #101, opened 2018-10-11
Add auto unseal test

Etiene:
* Adds test for vault auto-unseal

### DIFF
--- a/test/vault_helpers.go
+++ b/test/vault_helpers.go
@@ -24,6 +24,13 @@ import (
 )
 
 const REPO_ROOT = "../"
+
+// This is the alias of a KMS key we have previously created that lives in the
+// AWS account where our CI tests run. We have one with the same alias in
+// every region. This key is necessary for the test of an Enterprise Vault feature
+// called auto unseal. If you wish to run test this locally, replace this with
+// the alias of an KMS key you already have on the AWS account you use for running
+// your tests or create a new one. Beware that creating an AWS KMS key costs money.
 const AUTO_UNSEAL_KMS_KEY_ALIAS = "dedicated-test-key"
 
 const ENV_VAR_AWS_REGION = "AWS_DEFAULT_REGION"
--- a/test/vault_helpers.go
+++ b/test/vault_helpers.go
@@ -24,6 +24,7 @@ import (
 )
 
 const REPO_ROOT = "../"
+const AUTO_UNSEAL_KMS_KEY_ALIAS = "dedicated-test-key"
 
 const ENV_VAR_AWS_REGION = "AWS_DEFAULT_REGION"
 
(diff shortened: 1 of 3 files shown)


test/vault_helpers.go
```diff
@@ -24,6 +24,7 @@ import (
 )
 
 const REPO_ROOT = "../"
+const AUTO_UNSEAL_KMS_KEY_ALIAS = "dedicated-test-key"
```
brikis98: Add a comment explaining what this is, both so we know what this is, and open source community folks know to update this.

test/vault_helpers.go
```diff
@@ -412,6 +458,25 @@ func deployCluster(t *testing.T, examplesDir string, uniqueId string, terraformV
...
+	establishConnectionToCluster(t, cluster)
+	waitForVaultToBoot(t, cluster)
+
+	logger.Logf(t, "Initializing the cluster")
+	ssh.CheckSshCommand(t, cluster.Leader, "vault operator init")
+	assertStatus(t, cluster.Leader, Leader)
+
+	assertStatus(t, cluster.Standby1, Sealed)
+	restartVault(t, cluster.Standby1)
+	assertStatus(t, cluster.Standby1, Standby)
```
brikis98: Just to make sure I understand correctly... When Vault boots the very first time, it is not initialized and therefore sealed. Then you run `init`... And that unseals the leader... But you have to reboot the other nodes to unseal them? There's no way for those follower nodes to auto unseal after an init?

Etiene: Yes, that's correct. So I went ahead and manually tested and when follower nodes join the cluster afterwards they are unsealed. If they joined before init, they are sealed and they have to rejoin. Would it be possible to launch just the leader, init, and then add the standby nodes? Such adding a delay between leader and others? Or creating a cluster of 1, init, then increase the size of the asg group?

Etiene: I tested the later option, increasing the size of the cluster size and that works like a charm. New nodes boot unsealed. I guess we could just recommend that?

brikis98: Neat idea! I'm not sure what the official recommendation is, but scaling up the cluster seems like one reasonable way of doing it.

The other option is that you identify a "rally point" node that will initialize the cluster. All other nodes will ping the rally point and wait for it to boot and run `init` before booting their own nodes and auto-unsealing. This might be a nicer user experience. Here's how we identify a rally point in the couchbase code: https://github.com/gruntwork-io/terraform-aws-couchbase/blob/master/modules/couchbase-commons/couchbase-common.sh#L263

brikis98: Oh, right, as you pointed out on Slack, the rally point approach still means running `init` automatically on the server, which means there's no easy way to save the root token. OK, starting with a cluster of size one, running `init` manually, saving the root token, and then scaling up to 3 seems like a reasonable approach. As long as we clearly document that, I'm good 👍

Etiene:
Notify:
@infosecgithub
@bgt101

PR #105, opened 2018-10-18
Improve auto unseal docs

Etiene:


examples/vault-auto-unseal/README.md
```diff
@@ -56,6 +55,49 @@ we strongly recommend deploying the Vault cluster into the private subnets of a
...
+are enough parts to reconstruct the master key. This is done on different machines in the
+vault cluster for better security. Vault remains unsealed until it reboots or until
+someone manually reseals it.
+
+### Auto-unseal
+
+Vault Enterprise has a feature that allows automatic unsealing via Amazon KMS. It
+allows operators to delegate the unsealing process to AWS, which is useful for failure
+situations or creation of ephemeral clusters. This process uses an AWS KMS key as
+a [seal wrap][seal_wrap] mechanism: it encrypts and decrypts Vault's master key.
```
josh-padnick: > which is useful for failure situations or creation of ephemeral clusters.

Can you elaborate on these use cases? What specific problem(s) does auto-unseal solve?

josh-padnick: Also, can you make it more clear that AWS KMS is used to encrypt/decrypt the entire Master Key and replaces Shamir's Secret Sharing?

examples/vault-auto-unseal/README.md
```diff
@@ -56,6 +55,49 @@ we strongly recommend deploying the Vault cluster into the private subnets of a
...
+ * `--auto-unseal-kms-key-id`: The key id of the AWS KMS key to be used
+ * `--auto-unseal-region`: The AWS region where the KMS key lives
+
+In this example, like in other examples, we execute `run-vault` at the [`user-data`
+script][user_data], which runs at the boot of every node in the Vault cluster. The
+`key-id` is passed to this script by Terraform, after Terraform reads this value from a
+data source through the key alias. This means that the AWS key has to be previously
+manually created and we are using Terraform just to find this resource, not to
+create it. It is important to notice that AWS KMS keys have a [cost][kms_pricing]
+per month per key, as well as an API usage cost.
```
josh-padnick: s/which runs at the boot of every node/which runs on boot for every node/

josh-padnick: Nice warning on the KMS key cost, but we should do one better. Can you include a paragraph with a bold **Billing Warning:** at the beginning that points out that every time you create a KMS key, you're charged $1 for the month, even if you immediately delete it? For that reason, users will want to be careful not to automatically create and destroy KMS keys frequently.

examples/vault-auto-unseal/README.md
```diff
@@ -56,6 +55,49 @@ we strongly recommend deploying the Vault cluster into the private subnets of a
   `vault write /sys/license "text=$LICENSE_KEY_TEXT"`. If you don't do that, Vault
   will re-seal after 30 minutes and you won't be able to write or read secrets anymore.
 
+### Seal
+
+Vault servers boot sealed, which means it can access its storage, but can't decrypt it.
+So you can't really do anything apart from unsealing it or checking the server status.
+When you unseal, it recreates the master key, which isn't stored anywhere and is
+used to read the stored decryption key, which then can decrypt the data, and then
+you can start performing other operations on Vault.
```
josh-padnick: You might make this explanation more clear by first stating that all data stored by Vault is encrypted with a Master Key which Vault only ever keeps in memory (technically, there are "intermediate keys" used to encrypt the data and it's the intermediate keys that are decrypted with the Master Key, but that detail doesn't matter here). When Vault first boots, it does not have the Master Key in memory, and therefore cannot decrypt its own data, limiting its functionality to checking the server status or writing the Master Key to memory. In such a state, Vault is "sealed."

Etiene:
Notify
@infosecgithub  
@bgt101

PR #109, opened 2018-11-01
Tests restructure

Etiene:
* Refactors tests but **does not yet fix all the underlying issues** that was making some of them fail, the issues caused by amis failing  to build seem to be gone though :)
* Bumps terraform version
* Uses Eugenes multiple artifact builder
* Uses Yori's log parser
* Makes local ami building faster


examples/vault-consul-ami/vault-consul.json
```diff
@@ -52,11 +52,11 @@
   }],
   "provisioners": [{
     "type": "shell",
-    "inline": ["mkdir /tmp/terraform-aws-vault"]
+    "inline": ["mkdir -p /tmp/terraform-aws-vault/modules"]
```
Etiene: This bit reduced my upload when building the amis when testing locally from ~650MB to ~200KB because it avoids uploading `.terraform` dirs and `test/vendor`

test/vault_auth_test.go
```diff
@@ -1,25 +1,93 @@
...
+// Test the Vault EC2 authentication example by:
+//
+// 1. Copying the code in this repo to a temp folder so tests on the Terraform code can run in parallel without the
+//    state files overwriting each other.
+// 2. Building the AMI in the vault-consul-ami example with the given build name
+// 3. Deploying that AMI using the example Terraform code setting an example secret
+// 4. Waiting for Vault to boot, then unsealing the server, creating a Vault Role to allow logins from instances with a specific EC2 property and writing the example secret
+// 5. Waiting for the client to login, read the secret and launch a simple web server with the contents read
+// 6. Making a request to the webserver started by the auth client
+func runVaultEC2AuthTest(t *testing.T, amiId string, sshUserName string) {
```
Etiene: Nothing changed here besides removing the step of building the ami. Everything shows as new because I moved it from file. All tests were on `vault_helper` and with time that had become a monstrosity

test/vault_auth_test.go
```diff
@@ -1,25 +1,93 @@
...
+// Test the Vault IAM authentication example by:
+//
+// 1. Copying the code in this repo to a temp folder so tests on the Terraform code can run in parallel without the
+//    state files overwriting each other.
+// 2. Building the AMI in the vault-consul-ami example with the given build name
+// 3. Deploying that AMI using the example Terraform code setting an example secret
+// 4. Waiting for Vault to boot, then unsealing the server, creating a Vault Role to allow logins from resources with a specific AWS IAM Role and writing the example secret
+// 5. Waiting for the client to login, read the secret and launch a simple web server with the contents read
+// 6. Making a request to the webserver started by the auth client
+func runVaultIAMAuthTest(t *testing.T, amiId string, sshUserName string) {
```
Etiene: Same as other comment

test/vault_cluster_enterprise_test.go
```diff
@@ -35,3 +37,129 @@ func getUrlFromEnv(t *testing.T) string {
...
+
+// Test the Vault auto unseal example by:
+//
+// 1. Copying the code in this repo to a temp folder so tests on the Terraform code can run in parallel without the
+//    state files overwriting each other.
+// 2. Building the AMI in the vault-consul-ami example with the given build name
+// 3. Deploying a cluster of 1 vault server using the example Terraform code
+// 4. Sshing into vault node to initialize the server and check that it booted unsealed
+// 5. Increasing the the cluster size to 3 and check that new nodes are unsealed when they boot and join the cluster
+func runVaultAutoUnsealTest(t *testing.T, amiId string, sshUserName string) {
```
Etiene: Same as other comment

test/vault_cluster_private_test.go
```diff
@@ -2,14 +2,39 @@ package test
...
+// Test the Vault private cluster example by:
+//
+// 1. Copy the code in this repo to a temp folder so tests on the Terraform code can run in parallel without the
+//    state files overwriting each other.
+// 2. Build the AMI in the vault-consul-ami example with the given build name
+// 3. Deploy that AMI using the example Terraform code
+// 4. SSH to a Vault node and initialize the Vault cluster
+// 5. SSH to each Vault node and unseal it
+// 6. SSH to a Vault node and make sure you can communicate with the nodes via Consul-managed DNS
+func runVaultPrivateClusterTest(t *testing.T, amiId string, sshUserName string) {
```
Etiene: Same as other comment

test/vault_cluster_public_test.go
```diff
@@ -1,14 +1,49 @@
...
+// Test the Vault public cluster example by:
+//
+// 1. Copy the code in this repo to a temp folder so tests on the Terraform code can run in parallel without the
+//    state files overwriting each other.
+// 2. Build the AMI in the vault-consul-ami example with the given build name
+// 3. Deploy that AMI using the example Terraform code
+// 4. SSH to a Vault node and initialize the Vault cluster
+// 5. SSH to each Vault node and unseal it
+// 6. Connect to the Vault cluster via the ELB
+func runVaultPublicClusterTest(t *testing.T, amiId string, sshUserName string) {
```
Etiene: Same as other comment

test/vault_cluster_s3_backend_test.go
```diff
@@ -1,15 +1,80 @@
...
+// Test the Vault with S3 Backend example by:
+//
+// 1. Copy the code in this repo to a temp folder so tests on the Terraform code can run in parallel without the
+//    state files overwriting each other.
+// 2. Build the AMI in the vault-consul-ami example with the given build name
+// 3. Deploy that AMI using the example Terraform code
+// 4. SSH to a Vault node and initialize the Vault cluster
+// 5. SSH to each Vault node and unseal it
+// 6. Connect to the Vault cluster via the ELB
+func runVaultWithS3BackendClusterTest(t *testing.T, amiId string, sshUserName string) {
```
Etiene: Same as other comment

test/vault_main_test.go
```diff
@@ -0,0 +1,116 @@
...
+	t.Run("group", func(t *testing.T) {
+		runTestsOnDifferentPlatforms(t, testCases, amiIds)
+	})
+
+}
+
+func runTestsOnDifferentPlatforms(t *testing.T, testCases []testCase, amiIds map[string]string) {
+	var amiId string
+	for _, testCase := range testCases {
+		testCase := testCase
```
Etiene: I don't understand this line! I copied this from Yori's example. I'd appreciate some help understanding this.

yorinasub17: This is necessary, because the variable `testCase` is defined and set outside the forloop. As such, it gets overwritten on each iteration of the forloop. This is fine if you don't have concurrent code in the loop, but in this case, because you have a `t.Parallel`, the `t.Run` completes before the test function exits, which means that the value of `testCase` might change by the time the scheduler gets back into the test function (since you could have executed the next loop iteration).
By doing this line, where you are redefining the variable scoped to within the for loop, all references to that variable within the loop is no longer bound to what the value of the variable outside the loop is. You are therefore, freezing the value to what it is for that loop within the scope of the block.

Etiene: ah! thanks :D

brikis98: We usually add a comment about this

PR #111, opened 2018-11-08
Test/intermittent failures

Etiene:
This PR attempts to address remaining issues causing the vault tests to fail intermittently.

Still ongoing issues:
* Still hitting limit of instances on some regions => Todo: raise tickets with AWS
* Very rarely we fail to check health of vault when using consul as backend with `
2018-11-09T14:38:08.106Z [ERROR] core: error checking health: error="failed to check for initialization: Unexpected response code: 500"` as seen on this [CI run](https://circleci.com/gh/hashicorp/terraform-aws-vault/202). This might be related to the consul agent not having properly joined the cluster. I decided not to investigate this further or fix it for the moment and instead just raise an issue for this to be dealth with later

Resolved:
* Http request for downloading consul sometimes fails => added retry on consul repo the same way it was done here
* S3backend test unable to access bucket fixed by creating a dependency on creating the buckets and policies first when they are there by using the resources ids & names on a tag of the cluster
* Error deleting IAM instance profile was only happening locally on my machine, I guess state went weird? it just stopped happening
* Hitting limit of instances on most regions fixed by splitting amis into different regions 
* Vault Auth tests serving "null" because of race condition fixed by adding a check and retrying
* Vault Auth server failling to enable aws auth fixed by retrying
* Packer build failure for downloading package fixed by adding a retry on curl
* Init and apply failure fixed by ~~retrying init~~ adding a retryable error message to terraform options
* Instance profile role limit reached fixed for the moment by manually deleting a bunch of invalid ones directly at the AWS account
* Autoscaling group creation failure fixed by adding lifecycle (went in on previous PR)
* Probably some other stuff I forgot
* **Added retries on a bunch of things everywhere to prevent further errors** 
* Added a bunch of logs to artifacts, even on tests that were not presenting issues because **why not**


examples/vault-ec2-auth/user-data-auth-client.sh
```diff
@@ -107,10 +123,9 @@ retry_login "$data" "https://vault.service.consul:8200/v1/auth/aws/login"
 token=$(echo $login_output | jq -r .auth.client_token)
 
 # And use the token to perform operations on vault such as reading a secret
-response=$(curl \
-  -H "X-Vault-Token: $token" \
-  -X GET \
-  https://vault.service.consul:8200/v1/secret/example_gruntwork)
+response=$(retry \
+  "curl -H 'X-Vault-Token: $token' -X GET https://vault.service.consul:8200/v1/secret/example_gruntwork" \
+  "Trying to read secret from vault")
```
Etiene: The other changes in this file are there just so this bit here is retried

examples/vault-ec2-auth/user-data-vault.sh
```diff
@@ -74,7 +90,10 @@ echo "$server_output" | head -n 3 | awk '{ print $4; }' | xargs -l /opt/vault/bi
 export VAULT_TOKEN=$(echo "$server_output" | head -n 7 | tail -n 1 | awk '{ print $4; }')
 
 # Enables AWS authentication
-/opt/vault/bin/vault auth enable aws
+# This is an http request, and sometimes fails, hence we retry
+retry \
+  "/opt/vault/bin/vault auth enable aws" \
+  "Trying to enable aws auth"
```
Etiene: Same as other comment. The other changes in this file are there just so this bit here is retried

examples/vault-s3-backend/variables.tf
```diff
@@ -52,7 +52,7 @@ variable "vault_instance_type" {
 
 variable "consul_instance_type" {
   description = "The type of EC2 Instance to run in the Consul ASG"
-  default     = "t2.nano"
+  default     = "t2.micro"
```
Etiene: I had switched them to nano on previous PR, decided to switch them back to micro because errors caused by consul not being available started popping up, I didnt get to investigate origin in details because this improved it

modules/install-vault/install-vault
```diff
@@ -206,8 +230,9 @@ function fetch_binary {
     download_url="https://releases.hashicorp.com/vault/${version}/vault_${version}_linux_amd64.zip"
   fi
 
-  log_info "Downloading Vault to $DOWNLOAD_PACKAGE_PATH"
-  curl -o "$DOWNLOAD_PACKAGE_PATH" "$download_url" --location --silent --fail --show-error
+  retry \
+    "curl -o '$DOWNLOAD_PACKAGE_PATH' '$download_url' --location --silent --fail --show-error" \
+    "Downloading Vault to $DOWNLOAD_PACKAGE_PATH"
```
Etiene: TODO: Do the same for the consul repo

test/vault_cluster_enterprise_test.go
```diff
@@ -118,8 +116,9 @@ func testAutoUnseal(t *testing.T, asgNameOutputVar string, sshUserName string, t
 	establishConnectionToCluster(t, initialCluster)
 	waitForVaultToBoot(t, initialCluster)
 
-	logger.Logf(t, "Initializing the cluster")
-	ssh.CheckSshCommand(t, initialCluster.Leader, "vault operator init")
+	retry.DoWithRetry(t, "Initializing the cluster", 10, 10*time.Second, func() (string, error) {
+		return ssh.CheckSshCommandE(t, initialCluster.Leader, "vault operator init")
```
Etiene: `terraform init` was sometimes failing to install aws provider because it just fails to make the http request. I discussed with Josh and he suggested this type of change becomes built-in from terratest, but we should see how it behaves here first.

test/vault_helpers.go
```diff
@@ -115,7 +111,10 @@ func deployCluster(t *testing.T, amiId string, examplesDir string, uniqueId stri
 		},
 	}
 	test_structure.SaveTerraformOptions(t, examplesDir, terraformOptions)
-	terraform.InitAndApply(t, terraformOptions)
+	retry.DoWithRetry(t, "Running terraform Init", 3, 10*time.Second, func() (string, error) {
+		return terraform.InitE(t, terraformOptions)
```
Etiene: `terraform init` was sometimes failing to install aws provider because it just fails to make the http request and download the package. I discussed with Josh and he suggested this type of change becomes built-in from terratest, but we should see how it behaves here first.

test/vault_main_test.go
```diff
@@ -57,45 +71,47 @@ var testCases = []testCase{
...
-
 		tlsCert := generateSelfSignedTlsCert(t)
 		saveTlsCert(t, WORK_DIR, tlsCert)
 
-		var identifierToOptions = map[string]*packer.Options{
-			"vaultEnterpriseUbuntu":      composeAmiOptions(t, AMI_EXAMPLE_PATH, "ubuntu16-ami", tlsCert, awsRegion, getUrlFromEnv(t)),
-			"vaultEnterpriseAmazonLinux": composeAmiOptions(t, AMI_EXAMPLE_PATH, "amazon-linux-ami", tlsCert, awsRegion, getUrlFromEnv(t)),
-			"vaultUbuntu":                composeAmiOptions(t, AMI_EXAMPLE_PATH, "ubuntu16-ami", tlsCert, awsRegion, ""),
-			"vaultAmazonLinux":           composeAmiOptions(t, AMI_EXAMPLE_PATH, "amazon-linux-ami", tlsCert, awsRegion, ""),
+		amisPackerOptions := map[string]*packer.Options{}
```
Etiene: Instead of having all amis in the same region, each gets a random region now

test/vault_cluster_auth_test.go
```diff
@@ -89,5 +106,30 @@ func testRequestSecret(t *testing.T, terraformOptions *terraform.Options, expect
 	instanceIP := terraform.Output(t, terraformOptions, OUTPUT_AUTH_CLIENT_IP)
 	url := fmt.Sprintf("http://%s:%s", instanceIP, "8080")
 
-	http_helper.HttpGetWithRetry(t, url, 200, expectedResponse, 60, 10*time.Second)
+	http_helper.HttpGetWithRetry(t, url, 200, expectedResponse, 30, 10*time.Second)
+}
+
+func getSyslogs(t *testing.T, terraformOptions *terraform.Options, amiId string, awsRegion string, testName string) {
```
brikis98: FWIW, the file fetching stuff Eugene added is even more powerful: https://github.com/gruntwork-io/terratest/blob/master/modules/aws/ec2-files.go#L119. You can use it to grab syslog, vault logs, and config files from the ASG and write them somewhere on local disk to make debugging easier.

test/vault_main_test.go
brikis98: Is this to avoid limits in a region?

Etiene: yes

examples/vault-iam-auth/user-data-vault.sh
```diff
@@ -62,6 +78,10 @@ retry_init
 # existing unseal keys shares. See "vault operator rekey" for more information.
 # ==========================================================================
 
+
+# This is here only for debugging
+echo $server_output
```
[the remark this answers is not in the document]
Etiene: Not really

test/vault_cluster_auth_test.go
Etiene: Yes, but the client auth instance is not in asg

test/vault_helpers.go
brikis98: BTW, +1 on building this into Terratest (perhaps with a flag to optionally disable it), as this is definitely a common source of (transient) errors.

examples/vault-iam-auth/user-data-auth-client.sh
```diff
@@ -90,10 +94,10 @@ login_output=$(retry \
 token=$(echo $login_output | jq -r .auth.client_token)
 
 # And use the token to perform operations on vault such as reading a secret
-response=$(curl \
-  -H "X-Vault-Token: $token" \
-  -X GET \
-  https://vault.service.consul:8200/v1/secret/example_gruntwork)
+# These is being retried because race conditions were causing this to come up null sometimes
```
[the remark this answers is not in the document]
Etiene: I could if I knew 😂 I mean, I just assumed the issue was due to a race condition since both the cluster & the client are being launched at the same time and sometimes the client would complain about things such as "role doesnt exist yet"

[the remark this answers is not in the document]
brikis98: That exact thing may be worth adding as a comment

test/vault_cluster_auth_test.go
brikis98: There are methods to fetch from individual EC2 instances in that code too!

PR #115, opened 2018-11-29
allow vault servers by default

PascalBourdier:
In production, it is recommended to restrict access. Here's a clean way
for Vault servers to access the Consul cluster


Etiene:
Thanks for the PR! This would probably also be great to add to `examples/vault-cluster-private/main.tf`

PascalBourdier:
OK, I update my PR in the other examples

PR #117, opened 2018-12-05
Support configuring KMS API endpoint for auto-unseal

lawliet89:
- Removed `--auto-unseal-region` which was not used
- Added `--auto-unseal-kms-key-region` which was what was actually being used


Etiene:
Thanks for the PR and thanks for catching this!

PR #119, opened 2018-12-09
Fix "Private TLS Cert" Markdown markup

multani:
The numbered list describing the different steps was being reset due to the wrong bulleted list markup in between.

See difference between:

* [before the fix](https://github.com/hashicorp/terraform-aws-vault/blob/47c5fbc/modules/private-tls-cert/README.md#how-well-generate-a-tls-cert-for-private-services)
* [after the fix](https://github.com/hashicorp/terraform-aws-vault/blob/7706ca7f2c1a64efcccb9914542b768e50c0f5f8/modules/private-tls-cert/README.md#how-well-generate-a-tls-cert-for-private-services)

### DIFF
--- a/modules/private-tls-cert/README.md
+++ b/modules/private-tls-cert/README.md
@@ -187,20 +187,20 @@ this CA.
 So here's our strategy for issuing a TLS Cert for a private service:
 
 1. **Create our own CA.**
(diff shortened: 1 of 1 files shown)


Etiene:
Great catch! Thanks for the PR!

PR #120, opened 2018-12-10
Replace backtick in run-vault with single quote

briandbecker:
* Backticks causing unexpected behavior
* Removes backticks in favor of single quote to match other items in the
  list
* Credit to @BonzTM in #77 for initially finding/fixing; isolated to get
  fix merged

PR #122, opened 2018-12-17
Fix/run vault

Etiene:
* Does not configure unseal endpoint unless a value is explicitly passed
* Tested using vault 1.0 and turns out tests are still passing and auto unseal works fine using the open source version, with the same previous example :)


brikis98:
Is this a fix for what we saw in https://github.com/hashicorp/terraform-aws-vault/pull/120#issuecomment-445971334?

Etiene:
@brikis98 yup!

PR #123, opened 2018-12-18
Remove boto and pip installation from default installation

### LINKED ISSUE #116: Making Pip and Boto Installation Optional
AFAIK, pip and boto was added to the install script as part of the examples in https://github.com/hashicorp/terraform-aws-vault/pull/96

But pip and boto are strictly not required unless you are using the `vault-consul-ami` example. Shouldn't these installation be part of the example packer template for that example instead of being in the `install-vault` module? Alternatively, they can be optional.

More specific to my concern: I am using, by default, Python 3 on my Ubuntu AMIs. I am not sure what kind of issue installing pip and pip packages from Python 2 on top of Python 3 would bring. I am also not using `boto` in my AMI.

lawliet89:
Fixes #116

### DIFF
--- a/examples/vault-consul-ami/vault-consul.json
+++ b/examples/vault-consul-ami/vault-consul.json
@@ -98,7 +98,8 @@
   },{
     "type": "shell",
     "inline": [
-      "sudo apt-get install -y git"
+      "sudo apt-get install -y git python-pip",
+      "LC_ALL=C && sudo pip install boto3"
     ],
     "only": ["ubuntu16-ami"]
   },{
--- a/examples/vault-consul-ami/vault-consul.json
+++ b/examples/vault-consul-ami/vault-consul.json
@@ -7,6 +7,7 @@
     "consul_version": "1.3.1",
     "consul_download_url": "{{env `CONSUL_DOWNLOAD_URL`}}",
     "vault_download_url": "{{env `VAULT_DOWNLOAD_URL`}}",
+    "install_auth_signing_script": "true",
     "ca_public_key_path": null,
     "tls_public_key_path": null,
     "tls_private_key_path": null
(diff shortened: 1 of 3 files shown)


lawliet89:
`modules/install-vault/install-vault` was tested by building new AMIs and launching new instances successfully.

I didn't test the example.

brikis98:
We can't run automated tests automatically for external PRs for security reasons. Could you run those yourself? Or, if that's too much setup work, could you deploy the example with the new AMI and make sure the cluster comes up OK? Thank you!

lawliet89:
The AWS IAM auth script (and thus the Boto requirement) is not even used at all for starting up a Vault cluster.

Are you OK if I test using the [IAM Auth](https://github.com/hashicorp/terraform-aws-vault/tree/master/examples/vault-iam-auth) example instead?

I will see if I can make some more changes to only install Boto/Pip when needed for that example.

brikis98:
Sure! As long as you're putting the new AMI through its paces and it works, I'd be confident in merging this. Thanks!

lawliet89:
Did a couple of tests:

1. The [IAM Auth example](https://github.com/hashicorp/terraform-aws-vault/tree/master/examples/vault-iam-auth) which spins up a Vault and Consul cluster and an additional EC2 instance that authenticates via IAM worked.  (i.e. I was able to `curl http://<auth instance ip>:8080` and get back a secret I defined.)
1. I set `install_auth_signing_script` in the Packer template to anything other than `true` and I can confirm that Boto and Pip were not installed and the `sign-request.py` script is deleted.

```
ubuntu@ip-10-158-3-212:~$ sudo apt list --installed | grep python-pip

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

ubuntu@ip-10-158-3-212:~$ ls /opt/vault/scripts/
ubuntu@ip-10-158-3-212:~$ ls /tmp/
systemd-private-b9c55fb6e3aa41549d63e9cbf8003c6b-systemd-timesyncd.service-hN1JET
ubuntu@ip-10-158-3-212:~$ 

```

PR #124, opened 2018-12-19
Use double quotes for unseal endpoint

lawliet89:
Fixes `Error loading configuration from /opt/vault/config: error loading "/opt/vault/config/default.hcl": At 5:16: illegal char`

HCL does not support single quoting strings.


Etiene:
thanks for catching this!

PR #142, opened 2019-03-27
Switch from supervisord to systemd

### LINKED ISSUE #73: Use systemd instead of supervisord
On systems which have systemd available (more and more...), systemd might be used instead of supervisord. It saves runtime resources (no need to run supervisord and systemd at the same tie) and speeds up installation, hepful for autoscaling groups without pre-baked images.

edtan:
Closes #73 by switching to Amazon Linux 2 and changing to `systemd`.

The `vault.service` configuration was based off of the configuration listed at the [Vault Deployment Guide](https://learn.hashicorp.com/vault/operations/ops-deployment-guide).

Also, I've updated `README.md` to indicate that its been updated to Amazon Linux 2, but there is a link to the [latest public AMIs](https://github.com/hashicorp/terraform-aws-vault/blob/master/_docs/amazon-linux-ami-list.md).  That list will need to be updated with new Amazon Linux 2 AMIs if this PR is accepted.

I've also updated one of the examples to also install `pip` as it's still required by the `terraform-aws-consul` modules to install `supervisord`.  I'm looking into updating those modules to use `systemd` as well, so will remove the reference to `pip` once I have an PR in for those modules.

update: I submitted hashicorp/terraform-aws-consul#130, but it seems like `pip` and `boto3` is needed for the auth signing script example.  Thus, I've updated the PR to install them on Amazon Linux 2 as it seems like they're not already included.


Etiene:
Thanks for the PR! Im going to run the test suite

examples/vault-auto-unseal/README.md
```diff
@@ -49,7 +49,7 @@ even if you immediately delete it.
 1. Ssh to an instance in the vault cluster and run `vault operator init` to initialize
   the cluster, then `vault status` to check that it is unsealed. If you ssh to a
   different node in the cluster, you might have to restart Vault first with
-  `sudo supervisorctl restart vault` so it will rejoin the cluster and unseal.
+  `sudo systemctl restart vault` so it will rejoin the cluster and unseal.
```
brikis98: Would it be `restart vault` or `restart vault.service`?

examples/vault-consul-ami/vault-consul.json
```diff
@@ -113,9 +113,13 @@
   },{
     "type": "shell",
     "inline": [
-      "sudo yum install -y git"
+      "sudo yum install -y git",
+      "if [ '{{user `install_auth_signing_script`}}' = 'true' ]; then",
```
brikis98: Please use `[[ ]]` and `==`.

modules/run-vault/run-vault
```diff
@@ -226,32 +225,55 @@ EOF
...
+  log_info "Creating systemd config file to run Vault in $systemd_config_path"
+  cat > "$systemd_config_path" <<EOF
+[Unit]
+Description="HashiCorp Vault - A tool for managing secrets"
+Documentation=https://www.vaultproject.io/docs/
+Requires=network-online.target
+After=network-online.target
+ConditionFileNotEmpty=$config_path
+
+[Service]
```
brikis98: The `supervisord` config set the stdout and stderr log files explicitly. Where are the logs going with systemd? Are we assuming users will know to use `journalctl`? Should that be documented somewhere?

examples/vault-auto-unseal/README.md
edtan: Oops, I'll change it to `vault.service` for consistency.

modules/run-vault/run-vault
edtan: Added documentation and the `systemd-stdout` and `systemd-stderr` options

edtan:
I've made the suggested changes, and also removed the creation of the `/opt/vault/log` folder in `install-vault`.

Etiene:
Tests are failing to fetch logs (`test/vault_helpers.go:134`), I'll try to have a look at that, but let me know if you have any insights

edtan:
It might be because I removed `/opt/vault/log` in `install-vault`.  I'll get the tests set up on my machine this week and can look into it further.

Etiene:
Ok, thanks!

edtan:
Oh I see, thanks!  Since the logs are using `journalctl` now, I'll look into updating the tests to look there instead.

examples/vault-consul-ami/vault-consul.json
edtan: I initially changed all three occurrences to use `[[ ]]` and `==`, but discovered that this failed when building the Ubuntu image, getting an error (`[[: not found`).  I was able to work around this by specifying `#!/bin/bash -e` as the `inline_shebang` parameter.  This seems to only be necessary when building the Ubuntu image, so I did not specify it for the Amazon Linux 2 image.

edtan:
I've updated the tests so that it writes out the Vault logs to a file.  I ran the `TestVaultPrivateCluster` example locally and it passed for both the Ubuntu and Amazon Linux 2 images.  Note that the same logs are in the syslogs, but it might be more convenient for debugging purposes to still have the separate Vault logs.

brikis98:
@Etiene Could you take over getting this PR passing tests and merged from here?

PR #144, opened 2019-04-09
Fixes to the docs

yorinasub17:
I went through the exercise of trying out the root example and deploying a vault cluster. In the process I hit a few warnings about deprecated commands, so updated those in the docs and scripts. I also updated a few places in the quickstart that wasn't immediately obvious to me what to do.

1. Clear trailing whitespace where I can
1. Changed `vars.tf` to `variables.tf`
1. `vault-cluster-public` was renamed to `root-example`, but links weren't updated
1. `vault unseal` is deprecated in favor of `vault operator unseal`
1. `vault auth` for logging in is deprecated in favor of `vault login`
1. The example script fails if `ssh_key_name` is blank, so highlighted that in the root example docs.

Notes
- I could not get to use it in the end because I didn't configure TLS correctly. I didn't use a publicly accessible domain, so `vault login` failed.
- I didn't know the best way to address this, but the example script assumes ubuntu clusters. I ran the example using Amazon Linux 2 and when I was copy pasting the output I missed that it had used `ubuntu@` and was confused for a brief moment.

### DIFF
--- a/examples/root-example/README.md
+++ b/examples/root-example/README.md
@@ -1,25 +1,25 @@
-# Public Vault Cluster Example 
+# Public Vault Cluster Example
 
-This folder shows an example of Terraform code to deploy a [Vault](https://www.vaultproject.io/) cluster in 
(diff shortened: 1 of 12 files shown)


yorinasub17:
FYI I don't have commit rights to merge this.

PR #145, opened 2019-04-11
Add Vault agent configuration

### LINKED ISSUE #133: Installing as a Vault Agent
Using the Consul module it is possible to use the run-consul script to start consul in agent mode and allow vault to connect to localhost.

There doesn't appear to be a similar provision in this module, which would allow behaviour shown in the root README diagram in which vault agents connect to the vault servers from the client applications/instances. Am I missing something, or is there missing functionality?

edtan:
Closes #133.  This adds support for installing as a Vault agent.  Currently, only auto-auth is supported (this doesn't add in support for caching).

I had to update to Vault 1.1.0, since it seems there was some bug with the Vault agent not respecting the Vault address in the config file.

I added a Vault agent test, which was mostly copy+pasted from the IAM auth test, with the only change being the use of Vault agent instead of hitting the Vault API using `curl`.

For testing, I only ran the Vault agent test using the Amazon Linux 2 image, and the EC2 Auth test using both the Amazon Linux 2 and Ubuntu images.

### DIFF
--- a/examples/vault-agent/user-data-auth-client.sh
+++ b/examples/vault-agent/user-data-auth-client.sh
@@ -2,9 +2,10 @@
 # This script is meant to be run in the User Data of each EC2 Instance while it's booting. The script uses the
 # run-consul script to configure and start Consul in client mode. Note that this script assumes it's running in an AMI
 # built from the Packer template in examples/vault-consul-ami/vault-consul.json.
-# It then uses a python script to create a signed request to the AWS STS API and sends this encrypted request
-# to the Vault server when authenticating. After login, it reads a secret and exposes it in a simple web server for
-# test purposes.
+# It then uses Vault agent to automatically authenticate to the Vault server. After login, Vault agent writes the
+# authentication token to a file location, which you can use for your applications.  Note that by default, only the `vault`
+# user has access to the file, so you may need to grant the appropriate permissions to your application.
+# Finally, this script reads a secret and exposes it in a simple web server for test purposes.
 
 set -e
 
--- a/examples/vault-agent/user-data-auth-client.sh
+++ b/examples/vault-agent/user-data-auth-client.sh
@@ -52,8 +52,9 @@ function retry {
 # Start the Vault agent
 /opt/vault/bin/run-vault --agent --agent-auth-type iam --agent-auth-role "${example_role_name}"
 
-# Retry in case the Vault server is still booting and unsealing
-# Or in case run-consul running on the background didn't finish yet
+# Retry and wait for the Vault Agent to write the token out to a file.  This could be
+# because the Vault server is still booting and unsealing, or because run-consul
+# running on the background didn't finish yet
 retry \
   "[[ -s /opt/vault/data/vault-token ]] && echo 'vault token file created'" \
   "waiting for Vault agent to write out token to sink"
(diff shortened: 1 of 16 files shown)


examples/vault-agent/user-data-auth-client.sh
```diff
@@ -0,0 +1,77 @@
...
+# These variables are passed in via Terraform template interpolation
+/opt/consul/bin/run-consul --client --cluster-tag-key "${consul_cluster_tag_key}" --cluster-tag-value "${consul_cluster_tag_value}"
+
+# Start the Vault agent
+/opt/vault/bin/run-vault --agent --agent-auth-type iam --agent-auth-role "${example_role_name}"
+
+# Retry in case the Vault server is still booting and unsealing
+# Or in case run-consul running on the background didn't finish yet
+retry \
+  "[[ -s /opt/vault/data/vault-token ]] && echo 'vault token file created'" \
```
brikis98: Does the vault agent write this token? If so, could you update the comment to explain that?

modules/run-vault/run-vault
```diff
@@ -235,13 +321,23 @@ function generate_systemd_config {
   local -r vault_systemd_stdout="$5"
   local -r vault_systemd_stderr="$6"
   local -r vault_user="$7"
+  local -r is_vault_agent="$8"
   local -r config_path="$config_dir/$VAULT_CONFIG_FILE"
 
+  local vault_description="HashiCorp Vault - A tool for managing secrets"
+  local vault_command="server"
+  local vault_config_file_or_dir=$vault_config_dir  # Vault agent requires single file, but server can accept config dir.
```
brikis98:
```suggestion
  local vault_config_file_or_dir="$vault_config_dir"  # Vault agent requires single file, but server can accept config dir.
```

modules/run-vault/run-vault
```diff
@@ -235,13 +321,23 @@ function generate_systemd_config {
...
+  if [[ "$is_vault_agent" == "true" ]]; then
+    vault_command="agent"
+    vault_description="HashiCorp Vault Agent"
+    vault_config_file_or_dir=$config_path
```
brikis98:
```suggestion
    vault_config_file_or_dir="$config_path"
```

modules/run-vault/run-vault
```diff
@@ -134,6 +151,75 @@ function vault_version_at_least {
...
+      type = "$auth_type"
+      role = "$auth_role"
+    }
+  }
+
+  sink "file" {
+    config = {
+      path = "$data_dir/$VAULT_TOKEN_FILE"
+    }
+  }
```
brikis98: Oh, I see, you've added that file specifically as a sink. OK, up above in the script that reads this sink, please add comments explaining that!

examples/vault-agent/user-data-auth-client.sh
edtan: Updated the comment to reflect this

modules/run-vault/run-vault
edtan: Updated in latest commit

modules/run-vault/run-vault
edtan: Updated in latest commit

modules/run-vault/run-vault
edtan: Added a comment in the script, as well as a description in the README.

brikis98:
Looks like tests passed. Merging now, thank you!

brikis98:
https://github.com/hashicorp/terraform-aws-vault/releases/tag/v0.12.1

edtan:
My pleasure!  Likewise, thank you for the quick review and merge!  I've updated my scripts to point to the latest release.

PR #154, opened 2019-06-26
Terraform 0.12.x Upgrade

mcalhoun:
This PR makes all modules and examples compatible with Terraform 0.12.x.

Please note that this change is BACKWARDS INCOMPATIBLE and you MUST use Terraform 0.12.x and above with the code once this has been merged.


examples/vault-agent/main.tf
```diff
@@ -13,7 +13,7 @@ terraform {
 resource "aws_instance" "example_auth_to_vault" {
   ami           = "${var.ami_id}"
```
Etiene: I'm confused, shouldn't the examples also be updated? So shouldn't this be 
`ami           = var.ami_id` now?

main.tf
```diff
@@ -57,15 +59,15 @@ module "vault_cluster" {
...
-  cluster_name  = "${var.vault_cluster_name}"
-  cluster_size  = "${var.vault_cluster_size}"
-  instance_type = "${var.vault_instance_type}"
+  cluster_name  = var.vault_cluster_name
+  cluster_size  = var.vault_cluster_size
+  instance_type = var.vault_instance_type
 
-  ami_id    = "${var.ami_id == "" ? data.aws_ami.vault_consul.image_id : var.ami_id}"
-  user_data = "${data.template_file.user_data_vault_cluster.rendered}"
+  ami_id    = var.ami_id == "" ? data.aws_ami.vault_consul.image_id : var.ami_id
```
Etiene: Shouldn't the default of `var.ami_id` be null now?

modules/private-tls-cert/main.tf
```diff
@@ -32,41 +33,44 @@ resource "tls_self_signed_cert" "ca" {
...
+  ca_key_algorithm   = tls_private_key.ca.algorithm
+  ca_private_key_pem = tls_private_key.ca.private_key_pem
+  ca_cert_pem        = tls_self_signed_cert.ca.cert_pem
 
-  validity_period_hours = "${var.validity_period_hours}"
-  allowed_uses          = ["${var.allowed_uses}"]
+  validity_period_hours = var.validity_period_hours
+  allowed_uses          = var.allowed_uses
 
+  # Store the certificate's public key in a file.
```
Etiene: double comment

modules/vault-cluster/main.tf
```diff
@@ -11,42 +11,63 @@ terraform {
...
-  }"]
+  tag {
+    key                 = var.cluster_tag_key
+    value               = var.cluster_name
+    propagate_at_launch = true
+  }
+
+  tag {
+    key                 = "using_s3_bucket_backend"
+    value               = element(concat(aws_iam_role_policy.vault_s3.*.name, list("")), 0)
```
Etiene: Shouldn't it be `[""]` instead of `list("")`?

main.tf
mcalhoun: 46b245452069037aabba9d49d873fe6306c0c306

modules/private-tls-cert/main.tf
mcalhoun: f0e7d8863ba1fc0ab0f67c9c46bc07a98ca3b0c8

modules/vault-cluster/main.tf
mcalhoun: Ah, yes, this is indeed deprecated in 0.12.x 

https://www.terraform.io/docs/configuration/functions/list.html

mcalhoun: 88aa91789c20222b84927551844f0176b69a312f

examples/vault-agent/main.tf
mcalhoun: Yes, I hadn't pushed those commits for some reason. 

bbe0579
7027e0a
bbdf145
0ba9b1a
68a022e
2217a13

modules/vault-cluster/main.tf
```diff
@@ -1,52 +1,73 @@
...
+    propagate_at_launch = true
+  }
+
+  tag {
+    key                 = "using_auto_unseal"
+    value               = element(concat(aws_iam_role_policy.vault_auto_unseal_kms.*.name, [""]), 0)
+    propagate_at_launch = true
+  }
+
+  dynamic "tag" {
```
andrewward: Had to modify this to keep my module invocation the same. Does that seem correct?

Sample code:
```  
cluster_extra_tags = [
    {
      key                 = "Environment"
      value               = var.environment
      propagate_at_launch = true
    }
  ]
```

My modification:
```  
dynamic "tag" {
    for_each = [for s in var.cluster_extra_tags: {
      key                 = s.key
      value               = s.value
      propagate_at_launch = s.propagate_at_launch
    }]

    content {
      key                 = tag.value.key
      value               = tag.value.value
      propagate_at_launch = tag.value.propagate_at_launch
    }
  }
```

PR #157, opened 2019-07-02
Fix syntax of dynamic tag in block in ASG resource

rjshep:
The tag iterator object has a key and value attribute, it’s the value that’s needed to extract the actual key/value/propagate_at_launch attributes from.

### DIFF
--- a/modules/vault-cluster/main.tf
+++ b/modules/vault-cluster/main.tf
@@ -62,9 +62,9 @@ resource "aws_autoscaling_group" "autoscaling_group" {
     for_each = var.cluster_extra_tags
 
     content {
-      key                 = tag.key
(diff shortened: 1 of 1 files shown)


brikis98:
Thanks for the PR! I wonder why Terraform's type system didn't catch this? Kicking off tests shortly...

brikis98:
Apologies for the delay! The tests failed the first time I ran them, and I didn't have time to dig into the reason until just now. I poked around and it looked like an intermittent test failure, so I re-ran the tests, and they passed. I'll merge this now. Thanks!

brikis98:
https://github.com/hashicorp/terraform-aws-vault/releases/tag/v0.13.1

PR #160, opened 2019-08-26
Add --s3-bucket-path option for run-vault script

adriananeci:
As a new _**path**_ parameter allowing selecting the path within a bucket for Vault data was added in vault 1.2.0 (https://github.com/hashicorp/vault/blob/master/CHANGELOG.md#120-july-30th-2019) a new parameter needs to be supported by run-vault script.
It is useful when you configure S3 as a storage backend and want to store vault data at a specific path in the S3 bucket.


brikis98:
@Etiene Could you review?

Etiene:
@brikis98 I did, but the tests have been failing so I wanted to investigate why before giving a response.

Etiene:
Hi @adriananeci I have issued a fix of our test suite and am rerunning the tests of your changes on the top of that fix. Thanks for the patience! :)

PR #166, opened 2019-10-11
add one prod ready example

pgold30:
# Vault auto unseal cluster private with kms and acl for consul with encryption

Tested with latest version of Vault 1.2.3 and consul 1.6.1

This example combines the following examples: Vaul auto unseal - Vault private cluster - Consul with encryption and also adds some stuff that we expect in a prod ready working state, mainly:

* Secure instances using security groups and remove [0.0.0.0/0]
* Enable ACL on consul and pass the token in a secure way using AWS Secrets manager
* Enable encryption on Consul (based on https://github.com/hashicorp/terraform-aws-consul/tree/master/examples/example-with-encryption)
* Auto generate gossip_encryption_key and store it in AWS Secrets manager
* Only allow ssh from Vpn server
* Creates an ELB in front of Consul instances
* Creates an ELB in front of Vault instances
* Removed the block for deploying cluster in default VPC and availability zones, this needs to be set explicitly to enforce security

This example use Terragrunt to solve some dependencies but can by used with terraform also.

Thanks @mkrull for the ACL token provision


brikis98:
@pgold30 Thanks for the PR! This is awesome work. However, a few gotchas:

1. This code includes references to a bunch of proprietary Gruntwork code. That means the examples won't work for many open source users, leading to lots of 404's, confusion, etc. We should only use open source code in these open source repos.
1. It looks like this code is meant to be used in the Gruntwork Reference Architecture: e.g., it pull sin VPC data via `terraform_remote_state`. While very handy, it again won't work for many open source users who aren't using the Reference Architecture, Terragrunt, etc.
1. Each example in the `/examples` folder needs a corresponding test in the `/test` folder. Note that creating a KMS master key for each test run will be a bit too expensive ($1 per run!), so we'll need some tweaks there.
1. Side note: there's something odd/ironic about using AWS Secrets Manager to bootstrap Vault 😬 

If you want to simplify this example to something any open source user can use and make it testable, we can put it in this repo. Alternatively, we are planning to create an alternative repo later this quarter for Gruntwork customers where we can put such production-ready examples that include proprietary code, so another option is to wait until that's ready, and move your PR that to that repo. Let me know what you'd prefer!

pgold30:
Hi @brikis98 thanks for your feedback! 
1- Indeed we use gruntwork , terragrunt and other stuff, didn't wanted to remove this because i assume that will be nice for people using the same stack to have a prod ready example, will remove this stuff then in a new commit.
2- Will remove those refs
3-  I dont know how many test we run but having a prod ready example seems appropiate , the issue here since to be in the test code more than in this approach , since the auto-unseal  example makes use of this also.
4- Indeed, but wanted to find a solution for bootstraping vault, and use acl and token in consul whitout hardcoding so this looks like a nice and simple solution to do this, let me know if you think there is a better way.

Will apply the corrections to make it open source here and will be happy to help when you make the specific Gruntwork prod-ready example

pgold30:
@brikis98  Please let me know if the comments make sense or if i need to modify some stuff

brikis98:
Apologies, am currently traveling. Will check this out as soon as I can!

examples/vault-prod-ready/README.md
```diff
@@ -0,0 +1,180 @@
+# Vault auto unseal cluster private with kms and acl for consul with encryption. PROD READY
+
+This example icombines the followings: Vault auto unseal - Vault private cluster - Consul with encryption and also adds some stuff that we expect in a prod ready working state, mainly:
```
brikis98:
```suggestion
This example combines the following: Vault auto unseal, Vault private cluster, and Consul with encryption and also adds a few items that we expect in a production Vault deployment, including:
```

examples/vault-prod-ready/README.md
```diff
@@ -0,0 +1,180 @@
...
+* Secure instances using security groups and remove [0.0.0.0/0]
+* Enable ACL on consul and pass the token in a secure way using AWS Secrets manager
+* Enable encryption on Consul (based on https://github.com/hashicorp/terraform-aws-consul/tree/master/examples/example-with-encryption)
+* Auto generate gossip_encryption_key and store it in AWS Secrets manager
+* Only allow ssh from Vpn server
```
brikis98: Not everyone who uses Vault will have a VPN server. Perhaps this should be a more generic list of security group IDs to whitelist? You could put the security group ID of a VPN server in that list, or a bastion host, or whatever else your organization is using.

examples/vault-prod-ready/README.md
```diff
@@ -0,0 +1,180 @@
...
+* Creates an ELB in front of Consul instances
+* Creates an ELB in front of Vault instances
```
brikis98: Why the ELBs?

examples/vault-prod-ready/README.md
```diff
@@ -0,0 +1,180 @@
...
+* Removed the block for deploying cluster in default VPC and availability zones, this needs to be set explicitly to enforce security
+
+This example use Terragrunt to solve some dependencies but can by used with terraform also.
```
brikis98: We probably should keep these examples to Terraform-only. Anything that works with Terraform works with Terragrunt, but not necessarily vice versa.

examples/vault-prod-ready/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+Thanks @mkrull for the ACL token provision
+
+# HOWTO
+* 1 - Run terragrunt
+* 2 - run `vault operator init` on one of the vault servers
+* 3 - To login in consul look for master token under aws secrets manager
```
brikis98: Not sure I understand what this is saying. Could you try to clarify the language a bit?

examples/vault-prod-ready/README.md
```diff
@@ -0,0 +1,180 @@
...
+This example use Terragrunt to solve some dependencies but can by used with terraform also.
+
+Thanks @mkrull for the ACL token provision
+
+# HOWTO
+* 1 - Run terragrunt
+* 2 - run `vault operator init` on one of the vault servers
+* 3 - To login in consul look for master token under aws secrets manager 
+
+# Vault auto unseal example
```
brikis98: You already have a `#` (an `h1`) up above, so this should be `##`... And perhaps it needs a more specific title, as you already mentioned vault auto unseal example at the top.

examples/vault-prod-ready/README.md
```diff
@@ -0,0 +1,180 @@
...
+which allows them to use the Consul server cluster for service discovery and thereby
+access Vault via DNS using the domain name `vault.service.consul`.
+
+For more info on how the Vault cluster works, check out the [vault-cluster][vault_cluster]
+documentation.
+
+**Note**: To keep this example as simple to deploy and test as possible, it deploys
+the Vault cluster into your default VPC and default subnets, all of which are publicly
+accessible. This is OK for learning and experimenting, but for production usage,
+we strongly recommend deploying the Vault cluster into the private subnets of a custom VPC.
```
brikis98: This is not accurate, right? Is this a copy/paste from the other auto unseal example?

examples/vault-prod-ready/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+**Billing Warning**: Every time you create a KMS key, you're charged $1 for the month,
```
brikis98: Because of this, all of our KMS tests take in a KMS key ID as an input variable. That way, we can create a single KMS key we use for testing, rather than a new one for each test run.

examples/vault-prod-ready/README.md
```diff
@@ -0,0 +1,180 @@
...
+even if you immediately delete it.
+
+ Private Vault Cluster Example
```
brikis98: Another heading? More copy/paste below?

examples/vault-prod-ready/README.md
brikis98: Ah, looks like you may be doing just that. You may want to explain that here.

examples/vault-prod-ready/main.tf
```diff
@@ -0,0 +1,534 @@
...
+}
+
+data "aws_kms_alias" "vault-example" {
+  name = "alias/${var.auto_unseal_kms_key_alias}"
+}
+
+# ----------------------------------------------------------------------------------------------------------------------
+# SECURITY GROUPS
+# ----------------------------------------------------------------------------------------------------------------------
+# Only allow vault to be accessed from the OpenVPN server
```
brikis98: See not above on VPN servers

examples/vault-prod-ready/main.tf
```diff
@@ -0,0 +1,534 @@
...
+resource "aws_security_group_rule" "allow_vault_inbound_ssh" {
+  type                     = "ingress"
+  from_port                = 22
+  to_port                  = 22
+  protocol                 = "tcp"
+  source_security_group_id = var.security_group_id
+  security_group_id        = module.vault_cluster.security_group_id
+}
+
+# Only allow consul to be accessed from the OpenVPN server
```
brikis98: This comment and a few below have not been updated

examples/vault-prod-ready/main.tf
```diff
@@ -0,0 +1,534 @@
...
+
+  cluster_name  = var.vault_cluster_name
+  cluster_size  = var.vault_cluster_size
+  instance_type = var.vault_instance_type
+
+  ami_id    = var.ami_id
+  user_data = data.template_file.user_data_vault_cluster.rendered
+
+  vpc_id     = var.vpc_id
+#  subnet_ids = var.vpc_subnet_ids
```
brikis98: Why commented out?

examples/vault-prod-ready/main.tf
```diff
@@ -0,0 +1,534 @@
...
+  iam_role_id = module.vault_cluster.iam_role_id
+}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# GIVE SSH-GRUNT PERMISSIONS TO TALK TO IAM
+# We add an IAM policy to Jenkins that allows ssh-grunt to make API calls to IAM to fetch IAM user and group data.
+# ---------------------------------------------------------------------------------------------------------------------
+
+module "ssh_grunt_policies" {
+  source = "git::git@github.com:gruntwork-io/module-security.git//modules/iam-policies?ref=v0.19.1"
```
brikis98: `module-security` is a private Gruntwork repo. It should not be used in this public example.

examples/vault-prod-ready/main.tf
```diff
@@ -0,0 +1,534 @@
...
+  role   = module.consul_cluster.iam_role_id
+  policy = module.ssh_grunt_policies.allow_access_to_other_accounts[0]
+}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# ADD IAM POLICY THAT ALLOWS CLOUDWATCH LOG AGGREGATION
+# ---------------------------------------------------------------------------------------------------------------------
+
+module "cloudwatch_log_aggregation" {
+  source      = "git::git@github.com:gruntwork-io/module-aws-monitoring.git//modules/logs/cloudwatch-log-aggregation-iam-policy?ref=v0.14.0"
```
brikis98: Same for `module-aws-monitoring`

examples/vault-prod-ready/main.tf
```diff
@@ -0,0 +1,534 @@
...
+# we open up the consul specific protocols and ports for consul traffic
+# ---------------------------------------------------------------------------------------------------------------------
+
+module "security_group_rules" {
+  source = "github.com/hashicorp/terraform-aws-consul.git//modules/consul-client-security-group-rules?ref=v0.7.0"
+
+  security_group_id = module.vault_cluster.security_group_id
+
+  # To make testing easier, we allow requests from any IP address here but in a production deployment, we *strongly*
+  # recommend you limit this to the IP address ranges of known, trusted servers inside your VPC.
```
brikis98: Comment not updated

examples/vault-prod-ready/main.tf
```diff
@@ -0,0 +1,534 @@
...
+
+  allowed_inbound_cidr_blocks        = []
+  allowed_inbound_security_group_ids = []
```
brikis98: Shouldn't requests be allowed from *somewhere*?

examples/vault-prod-ready/main.tf
```diff
@@ -0,0 +1,534 @@
...
+
+  # The EC2 Instances will use these tags to automatically discover each other and form a cluster
+  cluster_tag_key   = var.consul_cluster_tag_key
+  cluster_tag_value = var.consul_cluster_name
+
+  ami_id    = var.ami_id
+  user_data = data.template_file.user_data_consul.rendered
+
+  vpc_id     = var.vpc_id
+#  subnet_ids = var.vpc_subnet_ids
```
brikis98: Same question

examples/vault-prod-ready/main.tf
```diff
@@ -0,0 +1,534 @@
...
+
+  # To make testing easier, we allow Consul and SSH requests from any IP address here but in a production
+  # deployment, we strongly recommend you limit this to the IP address ranges of known, trusted servers inside your VPC.
```
brikis98: Out of date comment

examples/vault-prod-ready/main.tf
```diff
@@ -0,0 +1,534 @@
...
+  listener {
+    instance_port     = 8500
+    instance_protocol = "http"
+    lb_port           = 8500
+    lb_protocol       = "http"
+  }
+}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# CREATE A SECURITY GROUP THAT CONTROLS WHAT TRAFFIC CAN GO IN AND OUT OF THE ELB OF CONSUL
```
brikis98: Consul module should already create this...

examples/vault-prod-ready/main.tf
```diff
@@ -0,0 +1,534 @@
...
+  name_prefix = "${var.consul_cluster_name}-token"
+}
+
+# Random uuid used as master token
+resource "random_uuid" "consul_token" {}
+
+# Secret version updated with the random uuid
+resource "aws_secretsmanager_secret_version" "consul_token" {
+  secret_id     = aws_secretsmanager_secret.consul_token.id
+  secret_string = random_uuid.consul_token.result
```
brikis98: This results in the secret being stored in Terraform state, which is not great. Would be better to generate this secret out of band and to use an IAM role that gives your servers permissions to read the secret.

examples/vault-prod-ready/main.tf
```diff
@@ -0,0 +1,534 @@
...
+}
+
+# Secret that holds the gossip encryption key
+resource "aws_secretsmanager_secret" "gossip_encryption_key" {
+  name_prefix = "gossip_encryption_key"
+}
+
+# Random uuid used as gossip encryption key
+resource "random_string" "gossip_encryption_key" {
```
brikis98: This too ends up stored in state. Again, would be better to add it to secrets manager out of band.

examples/vault-prod-ready/main.tf
```diff
@@ -0,0 +1,534 @@
...
+# Secret version updated with the random uuid
+resource "aws_secretsmanager_secret_version" "gossip_encryption_key" {
+  secret_id     = aws_secretsmanager_secret.gossip_encryption_key.id
+  secret_string = base64encode(random_string.gossip_encryption_key.result)
+}
+
+# Policy to allow Consul to write the consul_token secret and gossip encryption key
+resource "aws_iam_policy" "secretsmanager_get_token" {
+  name   = var.consul_cluster_name
+  policy = <<EOF
```
brikis98: Please use the `aws_iam_policy` data source.

examples/vault-prod-ready/main.tf
```diff
@@ -0,0 +1,534 @@
...
+  policy_arn = aws_iam_policy.secretsmanager_get_token.arn
+}
+
+data "aws_caller_identity" "current" {}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# DEPLOY THE CLUSTERS IN THE DEFAULT VPC AND AVAILABILITY ZONES
+# Using the default VPC and subnets makes this example easy to run and test, but it means Consul and Vault are
+# accessible from the public Internet. In a production deployment, we strongly recommend deploying into a custom VPC
+# and private subnets.
```
brikis98: I thought the docs said you weren't using a default VPC for this example?

examples/vault-prod-ready/user-data-consul.sh
```diff
@@ -0,0 +1,33 @@
...
+# built from the Packer template in examples/vault-consul-ami/vault-consul.json.
+
+set -e
+
+# Send the log output from this script to user-data.log, syslog, and the console
+# From: https://alestic.com/2010/12/ec2-user-data-output/
+exec > >(tee /var/log/user-data.log|logger -t user-data -s 2>/dev/console) 2>&1
+
+# These variables are passed in via Terraform template interpolation
+#/opt/consul/bin/run-consul --server --cluster-tag-key "${consul_cluster_tag_key}" --cluster-tag-value "${consul_cluster_tag_value}"
```
brikis98: ?

examples/vault-prod-ready/README.md
pgold30: Could be any load balancer , to balance request and still have one record entry when having multiple instances?

examples/vault-prod-ready/README.md
pgold30: Agree

examples/vault-prod-ready/README.md
pgold30: Sure, will change it to make it more clear. But basically it says that you need to ssh vault and run that command to get the token and that the consul token will be stored under aws secret manager

examples/vault-prod-ready/README.md
pgold30: Indeed, removed.

examples/vault-prod-ready/main.tf
pgold30: Agree

pgold30:
Sorry just got the chance to read all your suggestions , next week i will make all the changes requested, regarding the first part we can change prod ready for something like vault consul auto unseal with encryption or something similar. 
I can also write a guide on GW on how to deploy this in a prod way if you think thats the better approach since if not will need to remove some modules from here, can do both.
Thanks for your time regarding this PR , i will make it cleaner and address the changes next week.

PR #170, opened 2019-11-18
Pass IAM Instance profile ARN

mazzy89:
Signed-off-by: Salvatore Mazzarino <dev@mazzarino.cz>

This PR pass in as variable `iam_instance_profile_arn` to let the consumer be able to define the ARN of an instance profile managed externally of the lifecycle of the vault module.

If the user passes in the IAM Instance Profile ARN then all the policies that must be associated with the IAM roles will be managed by the user externally.

In some organizations, all IAM resources are not managed along with the TF modules but externally.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -66,8 +66,6 @@ module "vault_cluster" {
   ami_id    = var.ami_id == null ? data.aws_ami.vault_consul.image_id : var.ami_id
   user_data = data.template_file.user_data_vault_cluster.rendered
 
-  iam_instance_profile_arn = var.vault_iam_instance_profile_arn
-
   vpc_id     = data.aws_vpc.default.id
   subnet_ids = data.aws_subnet_ids.default.ids
 
--- a/main.tf
+++ b/main.tf
@@ -66,6 +66,8 @@ module "vault_cluster" {
   ami_id    = var.ami_id == null ? data.aws_ami.vault_consul.image_id : var.ami_id
   user_data = data.template_file.user_data_vault_cluster.rendered
 
+  iam_instance_profile_arn = var.vault_iam_instance_profile_arn
+
   vpc_id     = data.aws_vpc.default.id
   subnet_ids = data.aws_subnet_ids.default.ids
 
--- a/modules/vault-cluster/README.md
+++ b/modules/vault-cluster/README.md
@@ -313,7 +313,7 @@ Check out the [Security section](#security) for more details.
 ### IAM Role and Permissions
 
 Each EC2 Instance in the ASG has an [IAM Role](http://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles.html) attached.
-The IAM Role ARN is exported as an output variable so you can add custom permissions. If you require a deterministic IAM role ARN, the `iam_instance_profile_arn` variable can be set to your desired IAM Instance profile ARN.
+If you wish to use an IAM role and IAM instance profile created outside of this module, you can pass the instance profile ARN in via the `iam_instance_profile_arn` variable.
 
 
 ### S3 bucket (Optional)
--- a/modules/vault-cluster/README.md
+++ b/modules/vault-cluster/README.md
@@ -313,7 +313,7 @@ Check out the [Security section](#security) for more details.
 ### IAM Role and Permissions
 
 Each EC2 Instance in the ASG has an [IAM Role](http://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles.html) attached.
-The IAM Role ARN is exported as an output variable so you can add custom permissions.
+The IAM Role ARN is exported as an output variable so you can add custom permissions. If you require a deterministic IAM role ARN, the `iam_instance_profile_arn` variable can be set to your desired IAM Instance profile ARN.
 
 
 ### S3 bucket (Optional)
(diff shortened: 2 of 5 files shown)


mazzy89:
this PR should help to move on also with this https://github.com/hashicorp/terraform-aws-vault/pull/170

modules/vault-cluster/README.md
```diff
@@ -313,7 +313,7 @@ Check out the [Security section](#security) for more details.
 ### IAM Role and Permissions
 
 Each EC2 Instance in the ASG has an [IAM Role](http://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles.html) attached.
-The IAM Role ARN is exported as an output variable so you can add custom permissions.
+The IAM Role ARN is exported as an output variable so you can add custom permissions. If you require a deterministic IAM role ARN, the `iam_instance_profile_arn` variable can be set to your desired IAM Instance profile ARN.
```
brikis98: I'm not sure "deterministic" is the right term here. Perhaps:

```suggestion
If you wish to use an IAM role and IAM instance profile created outside of this module, you can pass the instance profile ARN in via the `iam_instance_profile_arn` variable.
```

modules/vault-cluster/main.tf
```diff
@@ -240,7 +244,7 @@ data "aws_iam_policy_document" "instance_role" {
 }
 
 resource "aws_s3_bucket" "vault_storage" {
-  count         = var.enable_s3_backend ? 1 : 0
+  count         = var.enable_s3_backend && var.iam_instance_profile_arn == null ? 1 : 0
```
brikis98: What does the S3 bucket have to do with the instance profile?

modules/vault-cluster/main.tf
```diff
@@ -295,7 +299,7 @@ data "aws_iam_policy_document" "vault_s3" {
 }
 
 data "aws_iam_policy_document" "vault_auto_unseal_kms" {
-  count = var.enable_auto_unseal ? 1 : 0
+  count = var.enable_auto_unseal && var.iam_instance_profile_arn == null ? 1 : 0
```
brikis98: Same with all of these? It doesn't seem like they should be conditional upon a custom instance profile?

modules/vault-cluster/variables.tf
```diff
@@ -167,6 +167,11 @@ variable "health_check_grace_period" {
   default     = 300
 }
 
+variable "iam_instance_profile_arn" {
+  description = "The ARN of the IAM Instance profilee to use for the Vault instances'. Use this setting to ensure a deterministic IAM instance role ARN, otherwise the default will see terraform create a unique IAM instance profile name."
+  default     = ""
```
brikis98: Add a `type`. Set `default` to `null`.

modules/vault-cluster/variables.tf
```diff
@@ -167,6 +167,11 @@ variable "health_check_grace_period" {
   default     = 300
 }
 
+variable "iam_instance_profile_arn" {
+  description = "The ARN of the IAM Instance profilee to use for the Vault instances'. Use this setting to ensure a deterministic IAM instance role ARN, otherwise the default will see terraform create a unique IAM instance profile name."
```
brikis98:
```suggestion
  description = "The ARN of the IAM instance profile to use instead of having this module create one internally."
```

variables.tf
```diff
@@ -90,6 +90,13 @@ variable "vault_instance_type" {
   default     = "t2.micro"
 }
 
+variable "vault_iam_instance_profile_arn" {
+  description = "The ARN of the IAM Instance profilee to use for the Vault instances'. Use this setting to ensure a deterministic IAM instance role ARN, otherwise the default will see terraform create a unique IAM instance profile name."
```
brikis98:
```suggestion
  description = "The ARN of the IAM instance profile to use instead of having this module create one internally."
```

modules/vault-cluster/main.tf
```diff
@@ -203,6 +203,7 @@ module "security_group_rules" {
 # ---------------------------------------------------------------------------------------------------------------------
 
 resource "aws_iam_instance_profile" "instance_profile" {
+  count       = var.iam_instance_profile_arn == null ? 1 : 0
```
brikis98: Each item that you've added a `count` parameter to is now treated as a list. You need to search for all references to this resource and all the other resources with `count` below and update those references to (a) only reference the resource if it exists and (b) use a list syntax.

main.tf
```diff
@@ -66,6 +66,8 @@ module "vault_cluster" {
   ami_id    = var.ami_id == null ? data.aws_ami.vault_consul.image_id : var.ami_id
   user_data = data.template_file.user_data_vault_cluster.rendered
 
+  iam_instance_profile_arn = var.vault_iam_instance_profile_arn
```
brikis98: There's no reason to chat the code in the root, as that is _example_ code. You only need this update in the modules in the `/modules` folder.

mazzy89:
@brikis98 thank you for the fast reaction here. Let me explain my changes.

The reason why I flag out all resources is that when a user passes in the ARN of the instance profile then also he created and managed a role with all the policies so no need to create policies there.

modules/vault-cluster/main.tf
mazzy89: Yeah good point. thank you for the catch here. going to fix it.

modules/vault-cluster/main.tf
mazzy89: https://github.com/hashicorp/terraform-aws-vault/pull/170#issuecomment-555454004

mazzy89:
I've addressed all the requested changes but not tested locally. Going to run asap locally and check if it works.

PR #183, opened 2020-02-07
Reduce Packer flakiness by updating apt-get

dmattia:
I was seeing some flakiness, where some dependencies like `awscli` were failing to download on the image after a `packer build`

Updating `apt-get` seemed to fix the flakiness for me

### DIFF
--- a/examples/vault-consul-ami/vault-consul.json
+++ b/examples/vault-consul-ami/vault-consul.json
@@ -74,6 +74,11 @@
     "type": "shell",
     "inline": ["mkdir -p /tmp/terraform-aws-vault/modules"]
   },{
+    "type": "shell",
+    "inline": [
+      "sudo apt-get update"
+    ]
+  },{
     "type": "file",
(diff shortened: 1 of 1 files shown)


robmorgan:
hi @dmattia we automatically run `sudo apt-get update -y` in the [`install-vault`](https://github.com/hashicorp/terraform-aws-vault/blob/master/modules/install-vault/install-vault#L121) script. How often was it failing to install the dependencies for you? Maybe it would be better to add retry logic instead vs adding a duplicate `apt-get` statement.

dmattia:
interesting, it was failing roughly 2/3rds of the time, and I believe I was still using install-vault.

Unfortunately, my config now looks quite different and I'll likely be unable to reproduce this issue again with which step was failing.

If nobody else has this issue or can reproduce it, this issue can likely be closed.

robmorgan:
Ok understood. I'll close the issue, but we can revisit it again in the future if anyone else experiences issues.

PR #189, opened 2020-03-09
Adding a flag to disable UI

stephencoe:
Adds a flag to disable UI when it is not needed. Enabled by default to preserve behavior


modules/run-vault/run-vault
```diff
@@ -463,6 +464,7 @@ function run {
   local auto_unseal_kms_key_id=""
   local auto_unseal_kms_key_region=""
   local auto_unseal_endpoint=""
+  local enable_ui="true"
```
brikis98: It doesn't look like you're letting the user set this flag anywhere? i.e., There's no new CLI arg exposed?

stephencoe: yes, good point! I was too quick opening the PR, I am looking at the tests for this now

PR #190, opened 2020-03-20
Storage backends

stephencoe:
# Description
Change `run-vault` to support dynamo backend as either storage *or* HA storage. 

## Why
The implementation of DynamoDB has limited support only allowing DynamoDB as both the `storage` and `ha_storage` backend. 

It is valid to use DynamoDB as either. These combinations should all be valid but not currently possible

| Storage  | HA Storage |
|----------|------------|
| S3       | ~DynamoDB~ |
| S3       | Consul     |
| DynamoDB | DynamoDB   |
| DynamoDB | ~Consul~   |
| Consul   | ~DynamoDB~ |
| Consul   | Consul     |

This change enables you to specify the `--storage-backend` as any of `s3|dynamodb|consul` and the `--ha-storage-backend` as `dynamodb|consul`. 

Consul is the default for both to maintain expected behavior

### DIFF
--- a/modules/run-vault/README.md
+++ b/modules/run-vault/README.md
@@ -72,10 +72,15 @@ The `run-vault` script accepts the following arguments:
 * `user` (optional): The user to run Vault as. Default is to use the owner of `config-dir`.
 * `skip-vault-config` (optional): If this flag is set, don't generate a Vault configuration file. This is useful if you
   have a custom configuration file and don't want to use any of of the default settings from `run-vault`.
-* `--enable-s3-backend` (optional): If this flag is set, an S3 backend will be enabled in addition to the HA Consul backend.
-* `--s3-bucket` (optional): Specifies the S3 bucket to use to store Vault data. Only used if `--enable-s3-backend` is set.
-* `--s3-bucket-path` (optional): Specifies the S3 bucket path to use to store Vault data. Default is `""`. Only used if `--enable-s3-backend` is set.
-* `--s3-bucket-region` (optional): Specifies the AWS region where `--s3-bucket` lives. Only used if `--enable-s3-backend` is set.
+* `--storage-backend` (optional): If this flag is set, the backend will be enabled in addition to the HA backend. Default is `consul`.
+* `--ha-storage-backend` (optional): If this flag is set, the HA backend will be enabled in addition to the storage backend. Default is `consul`.
+* `--s3-bucket` (optional): Specifies the S3 bucket to use to store Vault data. Only used if `--storage-backend` is set to `s3`.
+* `--s3-bucket-path` (optional): Specifies the S3 bucket path to use to store Vault data. Default is `""`. Only used if `--storage-backend` is set to `s3`.
+* `--s3-bucket-region` (optional): Specifies the AWS region where `--s3-bucket` lives. Only used if `--storage-backend` is set to `s3`.
+* `--dynamo-region` (optional): Specifies the AWS region where `--dynamo-table` lives. Only used if `--storage-backend` is set to `dynamodb`.
+* `--dynamo-table` (optional): Specifies the DynamoDB table name. Only used if `--storage-backend` is set to `dynamodb`.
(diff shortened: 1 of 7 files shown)


modules/run-vault/README.md
```diff
@@ -72,10 +72,15 @@ The `run-vault` script accepts the following arguments:
 * `user` (optional): The user to run Vault as. Default is to use the owner of `config-dir`.
 * `skip-vault-config` (optional): If this flag is set, don't generate a Vault configuration file. This is useful if you
   have a custom configuration file and don't want to use any of of the default settings from `run-vault`.
-* `--enable-s3-backend` (optional): If this flag is set, an S3 backend will be enabled in addition to the HA Consul backend.
-* `--s3-bucket` (optional): Specifies the S3 bucket to use to store Vault data. Only used if `--enable-s3-backend` is set.
-* `--s3-bucket-path` (optional): Specifies the S3 bucket path to use to store Vault data. Default is `""`. Only used if `--enable-s3-backend` is set.
-* `--s3-bucket-region` (optional): Specifies the AWS region where `--s3-bucket` lives. Only used if `--enable-s3-backend` is set.
+* `--storage-backend` (optional): If this flag is set, the backend will be enabled in addition to the HA backend. Default is `consul`.
```
brikis98: NIT: Should list all valid values (consul, s3, dynamo) for this param and the ones below.

modules/run-vault/run-vault
```diff
@@ -39,13 +39,15 @@ function print_usage {
...
   echo -e "  --skip-vault-config\tIf this flag is set, don't generate a Vault configuration file. Optional. Default is false."
-  echo -e "  --enable-s3-backend\tIf this flag is set, an S3 backend will be enabled in addition to the HA Consul backend. Default is false."
-  echo -e "  --s3-bucket\tSpecifies the S3 bucket to use to store Vault data. Only used if '--enable-s3-backend' is set."
-  echo -e "  --s3-bucket-path\tSpecifies the S3 bucket path to use to store Vault data. Only used if '--enable-s3-backend' is set."
-  echo -e "  --s3-bucket-region\tSpecifies the AWS region where '--s3-bucket' lives. Only used if '--enable-s3-backend' is set."
-  echo -e "  --enable-dynamo-backend\tIf this flag is set, DynamoDB will be enabled as the backend storage (HA)"
-  echo -e "  --dynamo-region\tSpecifies the AWS region where --dynamo-table lives.  Only used if '--enable-dynamo-backend is on'"
-  echo -e "  --dynamo--table\tSpecifies the DynamoDB table to use for HA Storage.  Only used if '--enable-dynamo-backend is on'"
+  echo -e "  --storage-backend\tStorage backend type to use for secrets. Default is consul"
```
brikis98: NIT: Same here. Should list all supported values for the backend params.

modules/run-vault/run-vault
```diff
@@ -282,58 +286,79 @@ listener "tcp" {
...
 )
-
-  local consul_storage_type="storage"
-  local dynamodb_storage_type="storage"
-  local s3_config=""
   local vault_storage_backend=""
-  if [[ "$enable_s3_backend" == "true" ]]; then
-    s3_config=$(cat <<EOF
+  local vault_ha_storage_backend=""
+  if [[ "$storage_backend" == "consul" ]]; then
```
brikis98: NIT: to keep this function from growing bigger and bigger, we should extract the code that sets the storage backend into a separate function that uses a case statement and returns the proper value by writing it to stdout. Same for HA storage.

modules/run-vault/run-vault
```diff
@@ -443,13 +468,15 @@ function run {
   local systemd_stderr=""
   local user=""
   local skip_vault_config="false"
-  local enable_s3_backend="false"
+  local storage_backend="consul"
+  local ha_storage_backend="consul"
```
brikis98: NIT: Store these values in `readonly` variables at the top. Use the variables here and in the docs.

PR #198, opened 2020-05-19
enable auto join for consul clients

Lucretius:
Consul seems to be the standard backend for this vault module, but cloud auto join is not supported currently because the IAM policies attached to the vault ec2 roles do not include the required permissions.  I've just copied these from the terraform-aws-consul module, though Consul itself claims it only needs the ec2:DescribeInstances permission set.

Tested as follows: Currently using this PR branch in place of the current master module in my own project, and cloud auto_join is working as expected.

### DIFF
--- a/modules/vault-cluster/main.tf
+++ b/modules/vault-cluster/main.tf
@@ -343,9 +343,9 @@ data "aws_iam_policy_document" "vault_dynamo" {
 }
 
 resource "aws_iam_role_policy" "vault_dynamo" {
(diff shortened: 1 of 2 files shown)


brikis98:
Thanks for the PR! I'm super booked this week, but will try to take a look in the next few days.

Lucretius:
Ah, actually thats what I was attempting to add, did not notice it was already in that other module.  I'll close this PR thanks.

Issue #196, opened 2020-04-12
Tests are unstable


brikis98:
They are failing intermittently. See #195 for details.

brikis98:
Should have been fixed by https://github.com/hashicorp/terraform-aws-vault/pull/195.

Issue #197, opened 2020-05-05
Vault Integrated Storage Support


toddmichael:
Hey guys,

First, thanks for this module.  I've been using it, along with your Consul module, to great effect for quite some time.  

With the release of `1.4.1` I'm considering giving the new [Vault Integrated Storage](https://www.vaultproject.io/docs/concepts/integrated-storage) feature a go to determine if/where it's an appropriate alternative to a separate Consul storage backend.

I perused open feature requests and didn't see anything planned yet for this so just thought I'd ask before plotting my next move.  If I head down this path before anyone else and there's interest, I'll gladly share any work I do to add support.  If there's no interest in adding this feature to this module, that's good to know as well, along with the logic behind the decision.  It may turn out that someone has already looked into this and determined that Consul (or S3) is still the better way to go in most production scenarios.  My logic has otherwise been quite simple: the feature was released via `1.4.0` so wait until at least `1.4.1` before kicking the tires

Thanks again for all the great work you contribute.  Cheers, be safe and be well.

Todd Michael

brikis98:
A big +1 on adding this. It's definitely still early days, so we probably shouldn't recommend it as the default setup yet and wait for the feature to bake a bit longer, but a PR to make it easy to use that functionality is very welcome!

os-virtualretail:
1.4.2 already released.

PR #195, opened 2020-04-12
Try to stabilize tests and update to AWS Provider 3.x

brikis98:
This is an attempt to stabilize the Vault tests. 

- Originally, they were failing due to intermittent issues downloading Terraform provider code; looks like the Terraform Registry is having some issues. Fixed with retries.
- Next, they were failing due to a lack of `t2.micro` in some regions. Fixed by excluding that region.

Now they are failing due to something on the server itself. [Example](https://circleci.com/gh/hashicorp/terraform-aws-vault/369?utm_campaign=build-failed&utm_medium=email&utm_source=notification). Typical output looks like this:

```
TestMainVaultCluster/group/TestVaultIAMAuthWithvaultOpenSourceOnUbuntu16Ami 2020-04-12T14:38:45Z command.go:158: [0m[1mmodule.consul_cluster.aws_autoscaling_group.autoscaling_group: Creation complete after 48s [id=consul-test-7j3F6A2020041214375729150000000c][0m[0m
TestMainVaultCluster/group/TestVaultIAMAuthWithvaultOpenSourceOnUbuntu16Ami 2020-04-12T14:38:45Z command.go:158: [0m[1m[32m
TestMainVaultCluster/group/TestVaultIAMAuthWithvaultOpenSourceOnUbuntu16Ami 2020-04-12T14:38:45Z command.go:158: Apply complete! Resources: 48 added, 0 changed, 0 destroyed.[0m
TestMainVaultCluster/group/TestVaultIAMAuthWithvaultOpenSourceOnUbuntu16Ami 2020-04-12T14:38:45Z command.go:158: [0m[1m[32m
TestMainVaultCluster/group/TestVaultIAMAuthWithvaultOpenSourceOnUbuntu16Ami 2020-04-12T14:38:45Z command.go:158: Outputs:
TestMainVaultCluster/group/TestVaultIAMAuthWithvaultOpenSourceOnUbuntu16Ami 2020-04-12T14:38:45Z command.go:158: 
TestMainVaultCluster/group/TestVaultIAMAuthWithvaultOpenSourceOnUbuntu16Ami 2020-04-12T14:38:45Z command.go:158: asg_name_consul_cluster = consul-test-7j3F6A2020041214375729150000000c
TestMainVaultCluster/group/TestVaultIAMAuthWithvaultOpenSourceOnUbuntu16Ami 2020-04-12T14:38:45Z command.go:158: asg_name_vault_cluster = vault-test-7j3F6A2020041214375542740000000b
[...]

PR #205, opened 2020-07-30
Support raft for HA storage

Pondidum:
I saw issue #197 about adding Vault's Integrated Storage for HA, and wanted to try it.

I am not certain that using the `$instance_ip_address` is the best thing for the `node_id`, but wasn't sure what else to use (and it works...).

I added two flags: `--enable-raft-backend` and `--raft-dir`.  `--raft-dir` should work identically to the `--config-dir` type flags, and the default direction is created by the `install-vault` script too.

I don't see any tests to verify this, other than it is working in my test aws account.


brikis98:

Pondidum:
You're welcome!

One thing I have noticed with the raft ha storage is that if I stop Vault on a node, or terminate the node (gracefully), the Vault instance never leaves the raft list of peers when checked on other nodes.

brikis98 (commented):
OK, thank you for your patience! We merged fixes in #195. Could you rebase?

The code changes here look great 👍 

Would you be up for switching one of the examples in the `examples` folder to using Raft? Or alternatively, adding a new example + test? I want to make sure we have tests that are executing this code path.

Pondidum:
I might have some time to do that later this week - I have just done the rebase for now.

Thanks

jboero:
Is there an update on this?  @Pondidum

Pondidum:
This had completely slipped my mind (I switched to using the dynamodb backend for my deployment), but I think I will have some time over the next week - I have some demos on Wednesday, so it would be after that I think.

I'll set a reminder this time...

Pondidum:
@jboero I've added an example which uses the raft backend, but there is an issue, which I also hit when trying to write the test for it too:  The raft version needs the certificate on each machine to be valid for the machine's private IP, not just `127.0.0.1` and `vault.service.consul`, as otherwise the machines fail to cluster due to TLS errors.

I am not sure how to go about generating the right certificates ahead of time, as the private IP is unknown until the machines boot.

RicoToothless:
> @jboero I've added an example which uses the raft backend, but there is an issue, which I also hit when trying to write the test for it too: The raft version needs the certificate on each machine to be valid for the machine's private IP, not just `127.0.0.1` and `vault.service.consul`, as otherwise the machines fail to cluster due to TLS errors.
> 
> I am not sure how to go about generating the right certificates ahead of time, as the private IP is unknown until the machines boot.

I have a similar issue with integrating Vault plugin.

The plugin uses `api_add` (default is private IP) to communicate with the vault server.

jboero:
Yes it seems like this is a pickle.  We shouldn't be using direct IPs when possible.  Other than a nasty wildcard cert I'm not sure what the answer is.

PR #206, opened 2020-08-01
Edit default value for availability_zones and subnet_ids vars to null

rafaelremondes:
Related to [Issue 181](https://github.com/hashicorp/terraform-aws-consul/issues/181) in `terraform-aws-consul`.

Newest `provider-aws` version only allows us to set one of the 2 variables: `availability_zones and subnet_ids` when creating AutoScalingGroups.

As per @brikis98 suggestion, I am changing the default value of these 2 variables to null so they're no longer conflicting, allowing to set either of them when using the `vault_cluster` module.


brikis98:

brikis98:
Thanks for the PR! This was already done as part of the fixes in #195.

PR #208, opened 2020-08-06
Adding max_instance_lifetime to vault-cluster ASG

yaleman:
Handy feature for those of us who need it, it's set to `0` by default which is effectively the same as before the change, but give people options.

### DIFF
--- a/modules/vault-cluster/variables.tf
+++ b/modules/vault-cluster/variables.tf
@@ -157,6 +157,11 @@ variable "wait_for_capacity_timeout" {
   default     = "10m"
 }
 
+variable "max_instance_lifetime" {
+  description = "The maximum amount of time, in seconds, that an instance can be in service, values must be either equal to 0 or between 604800 and 31536000 seconds."
+  default = 0
+}
+
 variable "health_check_type" {
   description = "Controls how health checking is done. Must be one of EC2 or ELB."
   default     = "EC2"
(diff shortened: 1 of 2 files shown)


brikis98:

modules/vault-cluster/variables.tf
```diff
@@ -157,6 +157,11 @@ variable "wait_for_capacity_timeout" {
   default     = "10m"
 }
 
+variable "max_instance_lifetime" {
+  description = "The maximum amount of time, in seconds, that an instance can be in service, values must be either equal to 0 or between 604800 and 31536000 seconds."
```
brikis98, line 161:
```suggestion
  description = "The maximum amount of time, in seconds, that an instance can be in service, values must be either equal to 0 (which means no limit) or between 604800 and 31536000 seconds."
```

modules/vault-cluster/variables.tf
```diff
@@ -157,6 +157,11 @@ variable "wait_for_capacity_timeout" {
...
+  default = 0
```
brikis98, line 162: Please add `type = number`

brikis98 (commented):
OK, thank you for your patience! We merged fixes in https://github.com/hashicorp/terraform-aws-vault/pull/195. Could you rebase?

yaleman:
I don't have resources to work on this further, closing.

PR #211, opened 2020-08-15
Ignore changes to load balancers in vault-cluster ASG

andrew-womeldorf:
Fixes #210

PR #212, opened 2020-08-17
prevent vault-cluster from removing vault-elb resources

Vivalldi:
When using the cluster with an elb attached the ASG wants to detach the target group. This is due to the change identification cycle.


Vivalldi:
This is actually a duplicate of #211 . I can nuke this if wanted

brikis98:
Thanks for the PR! This does indeed look like a duplicate of #211, so closing this one.

PR #224, opened 2020-10-16
Update repo to work with TF 0.13

brikis98:
1. Update `required_version` to `0.12.26` everywhere.
1. Add pre-commit hooks.
1. Update version numbers in CI build.
1. Run `go fmt` and `terraform fmt`.
1. Remove updating of Markdown files when we publish AMIs. Too much work to maintain, not enough value.

PR #225, opened 2020-10-19
Fixed link to "Latest Amazon Linux 2 AMIs"

domenjesenovec:
Fixed link to "Latest Amazon Linux 2 AMIs" 

The link took me to the wrong page, I used the outdated AMI and then I spent about an hour wondering how come DynamoDB is not supported etc.


brikis98:
Thanks for the fix! Note that as of https://github.com/hashicorp/terraform-aws-vault/pull/224, we are actually going to stop updating these Markdown files with AMI lists, so I'm going to close this PR. Note that we will continue publishing the AMIs themselves; it's just the Markdown files that are going away. You can still find the AMI IDs by doing a search in the AWS Marketplace or programmatically using tools like the `aws_ami` data source in Terraform.

Issue #228, opened 2020-11-15
Vault service isn't registered in consul.  UI not available via vault.service.consul


queglay:
I've been having some trouble being able to access the UI via https://vault.service.consul/ui in a private subnet.

I may be wrong, but I believe the examples showing HA and also the Private subnet example do not register the vault service with consul.  So unless you are using an ELB, you wont have the vault.service.consul FQDN to utilise the vault UI.

I am very new to Vault, but I'd imagine this would be problematic for others trying to use a redirect for a private cluster, OIDC, or use the dnsmasq scripts.

I tried to search for where this configuration block for the service registration would be specified in the repo, but couldn't find it:
```
service_registration "consul"
```
https://www.vaultproject.io/docs/configuration/service-registration/consul

Is this being specified anywhere or should it be by default?  Or is there anywhere in the repo that the vault service is getting registered with consul and I'm missing it?

I also opened a thread on hashicorp discuss yesterday before I realised this might be extending to other functions in this repository as well. https://discuss.hashicorp.com/t/why-might-a-consul-client-not-be-able-to-access-vault-ui-at-https-vault-service-consul-ui/17660

Thanks if anyone can provide any clues!

queglay:
Although I probably shouldn't do it this way, I tested appending the info into the run-vault script, and so far so good, the dig command pulls up an answer, and the web browser does resolve at this address.  I'm not sure if more is required to eliminate the https certificate warnings though, I'd like to figure that out.

```
    vault_storage_backend=$(cat <<EOF
$consul_storage_type "consul" {
  address = "127.0.0.1:8500"
  path    = "vault/"
  scheme  = "http"
  service = "vault"
}
# HA settings
cluster_addr  = "https://$instance_ip_address:$cluster_port"
api_addr      = "$api_addr"

service_registration "consul" {
  address = "127.0.0.1:8500"
  service = "vault"
  scheme  = "http"
}
EOF
)
```

brikis98:
This is probably the same cause as https://github.com/hashicorp/terraform-aws-vault/issues/223.

brikis98:
IIRC, using Consul as a backend and specifying a `service` name, should result in Vault being registered with Consul: https://github.com/hashicorp/terraform-aws-vault/blob/master/modules/run-vault/run-vault#L323. But perhaps some behavior changed to break that?

brikis98:
The [service registration docs](https://www.vaultproject.io/docs/configuration/service-registration) even say:

[...]

PR #232, opened 2021-01-27
Fixing tests using consul for dns

anouarchattouna:
* Refactoring tests
    * Add a new test stage `initialize_unseal` useful when running multiple times the `validation` stage against the same cluster
    * Add `testVaultUsesConsulForDns` for all test cases
* Updating vault service configuration
    * Adding `service_registration` used when using a different storage backend like `s3`
    * Updating `systemd` configuration based on this [configure-systemd documentation](https://learn.hashicorp.com/tutorials/vault/deployment-guide\#step-3-configure-systemd)
* Bumping versions
    * bump `min_packer_version` -> `1.5.4`
    * bump `vault_version` -> `1.6.1`
    * bump `consul_version` -> `1.9.2`

### DIFF
--- a/modules/run-vault/run-vault
+++ b/modules/run-vault/run-vault
@@ -44,6 +44,7 @@ function print_usage {
   echo -e "  --s3-bucket\tSpecifies the S3 bucket to use to store Vault data. Only used if '--enable-s3-backend' is set."
   echo -e "  --s3-bucket-path\tSpecifies the S3 bucket path to use to store Vault data. Only used if '--enable-s3-backend' is set."
   echo -e "  --s3-bucket-region\tSpecifies the AWS region where '--s3-bucket' lives. Only used if '--enable-s3-backend' is set."
+  echo -e "  --consul-agent-service-registration-address\tSpecifies the address of the Consul agent to communicate with when using a different storage backend, in this case an S3 backend. Only used if '--enable-s3-backend' is set."
   echo -e "  --enable-dynamo-backend\tIf this flag is set, DynamoDB will be enabled as the backend storage (HA)"
   echo -e "  --dynamo-region\tSpecifies the AWS region where --dynamo-table lives.  Only used if '--enable-dynamo-backend is on'"
   echo -e "  --dynamo--table\tSpecifies the DynamoDB table to use for HA Storage.  Only used if '--enable-dynamo-backend is on'"
@@ -73,7 +74,7 @@ function print_usage {
   echo
   echo "Or"
   echo
-  echo "  run-vault --tls-cert-file /opt/vault/tls/vault.crt.pem --tls-key-file /opt/vault/tls/vault.key.pem --enable-s3-backend --s3-bucket my-vault-bucket --s3-bucket-region us-east-1"
+  echo "  run-vault --tls-cert-file /opt/vault/tls/vault.crt.pem --tls-key-file /opt/vault/tls/vault.key.pem --enable-s3-backend --s3-bucket my-vault-bucket --s3-bucket-region us-east-1 --consul-agent-service-registration-address 127.0.0.1:8500"
 }
 
 function log {
@@ -237,13 +238,14 @@ function generate_vault_config {
   local -r s3_bucket="$9"
   local -r s3_bucket_path="${10}"
   local -r s3_bucket_region="${11}"
-  local -r enable_dynamo_backend="${12}"
-  local -r dynamo_region="${13}"
-  local -r dynamo_table="${14}"
+  local -r consul_agent_service_registration_address="${12}"
+  local -r enable_dynamo_backend="${13}"
+  local -r dynamo_region="${14}"
+  local -r dynamo_table="${15}"
   local -r enable_auto_unseal="${15}"
-  local -r auto_unseal_kms_key_id="${16}"
-  local -r auto_unseal_kms_key_region="${17}"
-  local -r auto_unseal_endpoint="${18}"
+  local -r auto_unseal_kms_key_id="${17}"
+  local -r auto_unseal_kms_key_region="${18}"
+  local -r auto_unseal_endpoint="${19}"
   local -r config_path="$config_dir/$VAULT_CONFIG_FILE"
 
   local instance_ip_address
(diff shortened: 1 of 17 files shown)


modules/run-vault/run-vault
```diff
@@ -299,9 +300,14 @@ EOF
 )
     consul_storage_type="ha_storage"
     dynamodb_storage_type="ha_storage"
+    service_registration=$(cat <<EOF
+service_registration "consul" {
+  address = "127.0.0.1:8500"
```
brikis98, line 305: Are we sure that it's always going to be a local Consul agent at that exact port? Or should these items be configurable?

modules/run-vault/run-vault
```diff
@@ -392,9 +401,12 @@ KillSignal=SIGINT
 Restart=on-failure
 RestartSec=5
 TimeoutStopSec=30
+StartLimitInterval=60
 StartLimitIntervalSec=60
 StartLimitBurst=3
 LimitNOFILE=65536
+LimitMEMLOCK=infinity
```
brikis98, line 408: I'm curious what are these settings tweaks doing?

test/vault_helpers.go
```diff
@@ -191,6 +191,24 @@ func initializeAndUnsealVaultCluster(t *testing.T, asgNameOutputVar string, sshU
...
+// Find the initialized and unsealed Vault cluster,
+// exit if cluster is not initialized and unsealed
+func getInitializedAndUnsealedVaultCluster(t *testing.T, asgNameOutputVar string, sshUserName string, terraformOptions *terraform.Options, awsRegion string, keyPair *aws.Ec2Keypair) VaultCluster {
+	cluster := findVaultClusterNodes(t, asgNameOutputVar, sshUserName, terraformOptions, awsRegion, keyPair)
+
+	establishConnectionToCluster(t, cluster)
+	isInitializedAndUnsealed, err := isVaultClusterInitializedAndUnsealed(t, cluster)
+	if err != nil {
+		t.Logf("Failed to check is vault cluster is already initialized and unsealed: %v", err)
+	}
```
brikis98, line 203: `require.NoError(t, err)` is shorter and does a better job of showing the error message in the logs.

brikis98 (commented):
Very nice, thanks! Just some minor NITs, and then I can kick off tests.

modules/run-vault/run-vault
anouarchattouna: This is mentioned [here](https://learn.hashicorp.com/tutorials/vault/deployment-guide#step-3-configure-systemd): `LimitNOFILE`, `LimitMEMLOCK` - Limit the number of file descriptors to the configured value and prevent memory being swapped to disk

modules/run-vault/run-vault
anouarchattouna: Indeed, better with configurable address

modules/run-vault/run-vault
brikis98: Ah, good to know, thanks!

modules/run-vault/run-vault
```diff
@@ -44,6 +44,7 @@ function print_usage {
   echo -e "  --s3-bucket\tSpecifies the S3 bucket to use to store Vault data. Only used if '--enable-s3-backend' is set."
   echo -e "  --s3-bucket-path\tSpecifies the S3 bucket path to use to store Vault data. Only used if '--enable-s3-backend' is set."
   echo -e "  --s3-bucket-region\tSpecifies the AWS region where '--s3-bucket' lives. Only used if '--enable-s3-backend' is set."
+  echo -e "  --consul-agent-service-registration-address\tSpecifies the address of the Consul agent to communicate with when using a different storage backend, in this case an S3 backend. Only used if '--enable-s3-backend' is set."
```
brikis98, line 47: Thanks for making this configurable! One suggestion though: let's make the default localhost:8500, so most users don't have to set it, though it'll still be available as an option if someone needs to override it.

brikis98 (approved):
This looks great, thank you! I'll kick off tests now.

modules/run-vault/run-vault
```diff
@@ -14,10 +14,12 @@ readonly DEFAULT_AGENT_AUTH_MOUNT_PATH="auth/aws"
 readonly DEFAULT_PORT=8200
 readonly DEFAULT_LOG_LEVEL="info"
 
+readonly DEFAULT_CONSUL_AGENT_SERVICE_REGISTRATION_ADDRESS="localhost:8500"
+
 readonly EC2_INSTANCE_METADATA_URL="http://169.254.169.254/latest/meta-data"
 
 readonly SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-readonly SCRIPT_NAME="$(basename "$0")"
+readonly SCRIPT_NAME="$(basename "${0}")"
```
brikis98, line 22: Why all the variable format changes? We don't usually do that when not necessary. Unless intentional, could you revert these? They make this diff very large.

brikis98 (commented):
Thanks! Just one comment and then I can kick off tests.

modules/run-vault/run-vault
anouarchattouna: I wanted to have the same format for all variables because there was a mix of usage:
```bash
  local -r s3_bucket="$9"	  
  local -r s3_bucket_path="${10}"
```
I'll revert anyway

brikis98 (approved):
Thanks! I'll kick off tests.

brikis98:
Tests passed! 🎉 

So does this imply there's no actual bug with https://github.com/hashicorp/terraform-aws-vault/issues/228 / Ubuntu 18.04? That it was just that the service wasn't being registered? Or did we need newer Consul / Vault versions?

anouarchattouna:
> Tests passed! 🎉
> 
> So does this imply there's no actual bug with #228 / Ubuntu 18.04? That it was just that the service wasn't being registered? Or did we need newer Consul / Vault versions?

Indeed, it resolves #228 by adding the service registration and using newer Consul / Vault versions

brikis98:
> > Tests passed! 🎉
> > So does this imply there's no actual bug with #228 / Ubuntu 18.04? That it was just that the service wasn't being registered? Or did we need newer Consul / Vault versions?
> 
> Indeed, it resolves #228 by adding the service registration and using newer Consul / Vault versions

Sorry, crazy week, just coming back to this now. That's great news, thank you! I'm going to merge and release this now.

brikis98:
https://github.com/hashicorp/terraform-aws-vault/releases/tag/v0.14.2

PR #234, opened 2021-03-19
Add missing '/' in source

ahamez:
To avoid the (granted, small) surprise that the commented source does not work when used as is.


brikis98 (approved):
Good catch. Thanks for the fix!

Issue #238, opened 2021-03-30
Add logging configuration to the ELB


dchernivetsky:
For security audit purposes, a logging configuration is required on the Load Balancer.
Even though configuration is straight passthrough, it is still useful to see and store the logs collected by the ELB itself.

brikis98:
Fixed in #238 and released in https://github.com/hashicorp/terraform-aws-vault/releases/tag/v0.14.3.

PR #239, opened 2021-03-30
addresses https://github.com/hashicorp/terraform-aws-vault/issues/238

dchernivetsky:


dchernivetsky:
#238

modules/vault-elb/main.tf
```diff
@@ -24,6 +24,17 @@ resource "aws_elb" "vault" {
   security_groups = [aws_security_group.vault.id]
   subnets         = var.subnet_ids
 
+  # optional access_logs creation  
+  dynamic "access_logs" {
+    for_each = length(keys(var.access_logs)) == 0 ? [] : [var.access_logs]
```
brikis98, line 29:
```suggestion
    for_each = var.access_logs == null ? [] : ["once"]
```

See the comment below for why I'm checking against `null`.

modules/vault-elb/variables.tf
```diff
@@ -25,6 +25,12 @@ variable "vault_asg_name" {
 # These parameters have reasonable defaults.
 # ---------------------------------------------------------------------------------------------------------------------
 
+variable "access_logs" {
+  description = "Map containing access logging configuration for load balancer."
+  type        = map(string)
```
brikis98, line 30: This can't be a `map(string)`, because then all values must be `string`, whereas the [`access_logs` block](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/elb) can contain a `bool` value (`enabled`) and an `int` value (`interval`). Due to limitations in Terraform's type system, that meets we lamely have to fall back to `any`. I'd recommend explaining this in a comment too. Also, you'll want to set the default to `null`.

modules/vault-elb/main.tf
```diff
@@ -24,6 +24,17 @@ resource "aws_elb" "vault" {
...
+
+    content {
+      enabled = lookup(access_logs.value, "enabled", lookup(access_logs.value, "bucket", null) != null)
+      bucket  = lookup(access_logs.value, "bucket", null)
+      bucket_prefix  = lookup(access_logs.value, "bucket_prefix", null)
```
brikis98, line 34: NIT: add `interval`.

modules/vault-elb/variables.tf
```diff
@@ -25,6 +25,12 @@ variable "vault_asg_name" {
 # These parameters have reasonable defaults.
 # ---------------------------------------------------------------------------------------------------------------------
 
+variable "access_logs" {
+  description = "Map containing access logging configuration for load balancer."
```
brikis98, line 29:
```suggestion
  description = "If specified, enables access logging for the ELB. This variable should be set to a map with the same keys and values as in the access_log block of the aws_elb resource (https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/elb), including bucket, bucket_prefix, interval, and enabled."
```

brikis98 (commented):
Thanks for the PR!

NIT: please run `terraform fmt` on the code.

dchernivetsky:
@brikis98 corrected the findings.

modules/vault-elb/main.tf
```diff
@@ -24,6 +24,18 @@ resource "aws_elb" "vault" {
...
 
+  # optional access_logs creation  
+  dynamic "access_logs" {
+    for_each = var.access_logs == null ? [] : ["once"]
+
+    content {
+      enabled       = lookup(access_logs.value, "enabled", lookup(access_logs.value, "bucket", null) != null)
+      bucket        = lookup(access_logs.value, "bucket", null)
+      bucket_prefix = lookup(access_logs.value, "bucket_prefix", null)
+      interval      = lookup(access_logs.value, "interval", 60)
```
brikis98, line 35: NIT: should `interval` also default to `null`?

modules/vault-elb/main.tf
```diff
@@ -24,6 +24,18 @@ resource "aws_elb" "vault" {
   security_groups = [aws_security_group.vault.id]
   subnets         = var.subnet_ids
 
+  # optional access_logs creation  
+  dynamic "access_logs" {
+    for_each = var.access_logs == null ? [] : ["once"]
+
+    content {
+      enabled       = lookup(access_logs.value, "enabled", lookup(access_logs.value, "bucket", null) != null)
```
brikis98, line 32: NIT: I think defaulting to `null` here would be OK:

```suggestion
      enabled       = lookup(access_logs.value, "enabled", lookup(access_logs.value, "bucket", null))
```

modules/vault-elb/variables.tf
```diff
@@ -25,6 +25,20 @@ variable "vault_asg_name" {
 # These parameters have reasonable defaults.
 # ---------------------------------------------------------------------------------------------------------------------
 
+# Due to terraform type limitations this variable is set to be a map of any.
```
brikis98, line 28: It actually has to be `any`, rather than `map(any)`. With `map(any)`, Terraform, sadly, would still require all the values to be of the same type, which won't work here. Until that's improved, we have to fall to the lowest common denominator of `type = any` and document the real type.

modules/vault-elb/variables.tf
```diff
@@ -25,6 +25,20 @@ variable "vault_asg_name" {
...
+# In reality it is a map of mixed parameters, some of them are booleans, some ints.
+#  access_logs = {
+#    enabled = true # mandatory
+#    bucket = "s3_bucket_name" # mandatory
+#    bucket_prefix = "alb" # mandatory
+#    interval = 60 # optional
```
brikis98, line 34: NIT: `fmt` this example for easier reading.

```suggestion
#    enabled = true # mandatory
#    bucket = "s3_bucket_name" # mandatory
#    bucket_prefix = "alb" # mandatory
#    interval = 60 # optional
```

modules/vault-elb/main.tf
dchernivetsky: I don't think interval should be null, when we can assign a sane value, which in this case we can, we shouldn't require the user to explicitly state what the interval is.

modules/vault-elb/main.tf
dchernivetsky: Well no right. enabled field takes a bool, so we should assign a bool. enabled = null is not the same as enabled = false. Tbh, i dont think enabled = null will work even.

modules/vault-elb/variables.tf
dchernivetsky: corrected

modules/vault-elb/main.tf
brikis98: [`enabled`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/elb#enabled) is listed as an optional param, with the default being `true`. If the user doesn't specify it, and we set it to `null`, we get this default. This allows us to fall back to Terraform's defaults, rather than having to provide our own.

modules/vault-elb/main.tf
brikis98: Same thing here: [`interval`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/elb#interval) is listed as an optional. If the user doesn't provide it, and we set it to `null`, we automatically get Terraform's default, rather than having to provide our own.

modules/vault-elb/variables.tf
brikis98: Whoops, I didn't mean to include the suggested formatting in there! But, as it's there, I've now updated to actually look as if it's had `fmt` run on it to make it easier to read.

modules/vault-elb/main.tf
dchernivetsky: corrected

dchernivetsky:
corrected all @brikis98

brikis98 (approved):
LGTM, thanks! I'll kick off tests now.

brikis98:
Tests passed! Merging now.

brikis98:
https://github.com/hashicorp/terraform-aws-vault/releases/tag/v0.14.3

dchernivetsky:
@brikis98 on second thought. I just tried to use 0.14.3 It doesnt appear that approach taken in the PR is valid. Specifically with ["once"] snippet.
```
Error: Invalid function argument

  on .terraform/modules/vault_cluster_elb/modules/vault-elb/main.tf line 32, in resource "aws_elb" "vault":
  32:       enabled       = lookup(access_logs.value, "enabled", lookup(access_logs.value, "bucket", null))
    |----------------
    | access_logs.value is "once"

Invalid value for "inputMap" parameter: lookup() requires a map as the first
argument.
```

Issue #236, opened 2021-03-30
When using S3 bucket for vault backend, can we destroy vault without destroying the bucket?


queglay:
In using the S3 backend option, I found that it was problematic to create the S3 bucket as part of this module.  

Once using a backend like S3, it becomes possible to destroy and deploy vault on a daily basis, but I don't want to destroy the bucket.  Would you accept a PR to use an existing bucket instead of creating one as the default workflow?

I'm suggesting it as a default just because I think it defeats the purpose of using the S3 backend and putting it in this module, since a terraform destroy on this module would be just like having no S3 backend at all.

Alternatively if this current implementation does have value the way it is for reasons I haven't considered... then could we use an optional var like 'use_existing_bucket=true' to utilise a data module to aquire an existing bucket instead of creating one?

brikis98:
Running `destroy` on a Vault cluster on a regular basis is not a normal use case, so I think the default as-is works fine. That said, a PR to allow providing an external S3 bucket is a reasonable alternative use case, and a PR to add that is welcome!

PR #240, opened 2021-04-02
Add option to use_existing_s3_bucket instead of creating one.

queglay:
If var.use_existing_s3_bucket is true, will acquire the bucket id and arn by the provided var.s3_bucket_name instead of creating a new bucket in this module.  This allows a user to avoid creating and destroying the S3 bucket repeatedly, and they can persist configuration from the s3 backend even after destroying this module.

### DIFF
--- a/modules/vault-cluster/main.tf
+++ b/modules/vault-cluster/main.tf
@@ -54,7 +54,7 @@ resource "aws_autoscaling_group" "autoscaling_group" {
 
   tag {
     key                 = "s3_bucket_id"
-    value               = element(concat(aws_s3_bucket.vault_storage.*.id, [""]), 0)
+    value               = local.s3_bucket_id
     propagate_at_launch = true
   }
 
@@ -268,8 +268,13 @@ data "aws_iam_policy_document" "instance_role" {
   }
 }
 
+locals {
+  s3_bucket_arn = var.use_existing_s3_bucket ? element(concat(data.aws_s3_bucket.vault_storage.*.arn, [""]), 0) : element(concat(aws_s3_bucket.vault_storage.*.id, [""]), 0)
+  s3_bucket_id = var.use_existing_s3_bucket ? element(concat(data.aws_s3_bucket.vault_storage.*.id, [""]), 0) : element(concat(aws_s3_bucket.vault_storage.*.id, [""]), 0)
+}
+
 resource "aws_s3_bucket" "vault_storage" {
-  count         = var.enable_s3_backend ? 1 : 0
+  count         = ( var.enable_s3_backend && var.use_existing_s3_bucket == false ) ? 1 : 0
   bucket        = var.s3_bucket_name
   force_destroy = var.force_destroy_s3_bucket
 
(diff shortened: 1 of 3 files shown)


queglay:
I'm just checking, did I create this pull request correctly?  It relates to the enhancement suggestion - https://github.com/hashicorp/terraform-aws-vault/issues/236

modules/vault-cluster/main.tf
```diff
@@ -268,8 +268,13 @@ data "aws_iam_policy_document" "instance_role" {
...
 }
 
+locals {
+  s3_bucket_arn = var.use_existing_s3_bucket ? element(concat(data.aws_s3_bucket.vault_storage.*.arn, [""]), 0) : element(concat(aws_s3_bucket.vault_storage.*.id, [""]), 0)
+  s3_bucket_id = var.use_existing_s3_bucket ? element(concat(data.aws_s3_bucket.vault_storage.*.id, [""]), 0) : element(concat(aws_s3_bucket.vault_storage.*.id, [""]), 0)
+}
+
 resource "aws_s3_bucket" "vault_storage" {
-  count         = var.enable_s3_backend ? 1 : 0
+  count         = ( var.enable_s3_backend && var.use_existing_s3_bucket == false ) ? 1 : 0
```
brikis98, line 277: NIT: please run `terraform fmt` on all the code changes in this PR; otherwise, our pre-commit hook will fail.

modules/vault-cluster/main.tf
```diff
@@ -268,8 +268,13 @@ data "aws_iam_policy_document" "instance_role" {
   }
 }
 
+locals {
+  s3_bucket_arn = var.use_existing_s3_bucket ? element(concat(data.aws_s3_bucket.vault_storage.*.arn, [""]), 0) : element(concat(aws_s3_bucket.vault_storage.*.id, [""]), 0)
+  s3_bucket_id = var.use_existing_s3_bucket ? element(concat(data.aws_s3_bucket.vault_storage.*.id, [""]), 0) : element(concat(aws_s3_bucket.vault_storage.*.id, [""]), 0)
```
brikis98, line 273: The `element(concat(..., [""]), 0)` workaround is fairly old code. Now that we're on newer versions of Terraform, we can switch to the following safer workaround:

```suggestion
  s3_bucket_arn = (
    length(aws_s3_bucket.vault_storage) > 0 
    ? aws_s3_bucket.vault_storage[0].arn 
    : length(data.aws_s3_bucket.vault_storage) > 0 ? data.aws_s3_bucket.vault_storage[0].arn : null
  )  
  s3_bucket_id = (
    length(aws_s3_bucket.vault_storage) > 0 
    ? aws_s3_bucket.vault_storage[0].id 
    : length(data.aws_s3_bucket.vault_storage) > 0 ? data.aws_s3_bucket.vault_storage[0].id : null
  )
```

brikis98 (commented):
Thanks for the PR!

modules/vault-cluster/main.tf
queglay: I must admit, the new method we should use is not easier to understand.  Thanks for pointing out how I should upgrade an old habit though!

queglay:
I'm wondering what other needs might need to be met for this to be approved?

queglay:
Is anything preventing this from being merged?

Issue #148, opened 2019-04-30
sign-request.py fails due to base64 encoding error


eedwards-sk:
see https://github.com/hashicorp/terraform-aws-vault/blob/master/examples/vault-consul-ami/auth/sign-request.py#L43

```
Traceback (most recent call last):
File "/opt/vault-login/sign-request.py", line 52, in <module>
vault_request = generate_vault_request()
File "/opt/vault-login/sign-request.py", line 34, in generate_vault_request
'iam_request_url': base64.b64encode(request.url),
File "/usr/local/lib/python3.6/base64.py", line 58, in b64encode
encoded = binascii.b2a_base64(s, newline=False)
TypeError: a bytes-like object is required, not 'str'
```

eedwards-sk:
This appears to be an issue with running against python 3.x

in Python 2.x, `base64.b64encode` takes a string object

in Python 3.x, it takes a bytes object

I'm not sure what the best approach is for making it compatible with both, but here's how I got it working in python 3.x:

```
    return {
        'iam_http_request_method': request.method,
        'iam_request_url':         base64.b64encode(bytes(request.url, 'utf-8')),
        'iam_request_body':        base64.b64encode(bytes(request.body, 'utf-8')),
        'iam_request_headers':     base64.b64encode(json.dumps(headers_to_go_style(dict(request.headers)))), # It's a CaseInsensitiveDict, which is not JSON-serializable
    }
```

also had to change `iteritems` to `items`

```
def headers_to_go_style(headers):
    retval = {}
    for k, v in headers.items():
        retval[k] = [v]
    return retval
```

probably some more changes too... I'll log them here as I find them

Etiene:
> I'm not sure what the best approach is for making it compatible with both
I think it's okay if it's py 3 only. PRs welcome! :)

yorinasub17:
What about adding a separate py3 version of the file (`sign-request-py3.py`)? Then users can pick and choose which one to include in their AMI when they are copying the examples.

I agree that we shouldn't try to make the existing file py2 + 3 compatible because without `six`, that will make the code littered with `if` statements.

eedwards-sk:
okay, it needed a few more tweaks, but here's a version for 3.x:

https://gist.github.com/eedwards-sk/161e98844db85919f6a85a3f7dce58f7

please feel free to toss into a PR or we could use this to figure out a 3.x/2.x compatible version

PR #243, opened 2021-04-16
Fixes issue #148 by adding python3 support to sign_request.py

bastianb:
An attempt to fix https://github.com/hashicorp/terraform-aws-vault/issues/148
Adds supports for python3 to sign_request.py


examples/vault-consul-ami/auth/sign-request.py
```diff
@@ -15,18 +15,28 @@
...
+        try:
+            value = v.decode()
+        except AttributeError:
+            value = v
+
+        if to_go_style:
+            value = [value]
+
+        retval[k] = value
     return retval
```
yorinasub17, line 37: I think using a single function for these two purposes is confusing. I would probably convert these function calls to inline dict comprehensions (which are 2 and 3 compatible) that do the things we want for each case.

